PR #20, opened 2021-09-08
Changed all Regexes to disallow spaces

RedingJeff:
Now the fields are really FIWARE safe. 

In two cases I shortened the regex, but as far as I saw (and the text stateted) these limitations were not needed for these fields. * is an allowed char.

### DIFF
--- a/filip/models/base.py
+++ b/filip/models/base.py
@@ -95,6 +95,14 @@ class FiwareRegex(str, Enum):
     Collection of Regex expression used to check if the value of a Pydantic
     field, can be used in the related Fiware field.
     """
-    standard = r"(^((?![?&#/ ])[\x00-\x7F])*$)"
+    _init_ = 'value __doc__'
+
+    standard = r"(^((?![?&#/ ])[\x00-\x7F])*$)", \
+               "Prevents any string that contains at least one of the " \
+               "symbols: ? & # / or a whitespace"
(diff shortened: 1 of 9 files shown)


tstorek (changes requested):
@RedingJeff 
Thanks for the pull request. Maybe this would be also a great opportunity to introduce a test for this to the changed models. Don't you think?

I think that * is part of the regex expression. Please, double check on e.g. https://www.regexpal.com/

RedingJeff:
I added tests for all the fields. 
To reduce the redundancy and improve maintainability, I grouped the two FiwareRegexes in a class togehter

I checked the simple regex again, the "*" is not needed, but the regex that should prevent specific strings was not correct. I corrected it.

filip/models/base.py
```diff
@@ -88,3 +88,13 @@ class FiwareHeader(BaseModel):
     class Config(BaseConfig):
         allow_population_by_field_name = True
         validate_assignment = True
+
```
tstorek, line 91: Almost there. Maybe we can use annotated enum here. Hence, we would alogn it with the other enums used in the library.

tstorek (changes requested):
Hi. Looks good. Please raise the version number by a minor release and write a small changelog comment as Bugfix. Easiest way is to search the project files for the current version number and and then adjust it :)

Issue #22, opened 2021-09-13
Post_subscription prevent duplicates


RedingJeff:
Prevent that post_subscription (ngsi_v2.cb.py) creates a duplicate entry.

The methode shall now check, if the exact same subscription already exists.
If yes, it makes a log entry and returns the ID of the already created subscription

tstorek:
closed by #23

PR #23, opened 2021-09-13
Issue#22 post subscription duplicates

RedingJeff:
Finished the Issue.

If now a subscription with the same subject or notification is posted as an already existing subscription. The existing subscription is updated instead (description, expiration_time,... )

I did not use __eq__ in Subscription as the used comparision is quite specific.

### DIFF
--- a/filip/clients/ngsi_v2/cb.py
+++ b/filip/clients/ngsi_v2/cb.py
@@ -885,13 +885,27 @@ class ContextBrokerClient(BaseHttpClient):
         Creates a new subscription. The subscription is represented by a
         Subscription object defined in filip.cb.models.
 
+        If the subscription already exists, the adding is prevented and the id
+        of the existing subscription is returned.
+
         Args:
-            subscription:
+            subscription: Subscription
 
         Returns:
-            object: 
+            str: Id of the created subscription
 
         """
+        existing_subscriptions = self.get_subscription_list()
+
+        sub_hash = str(subscription.dict(include={'subject', 'notification'}))
+        for ex_sub in existing_subscriptions:
+            if sub_hash ==\
+                    str(ex_sub.dict(include={'subject', 'notification'})):
+                self.logger.info("Subscription already exists, updated instead")
+                subscription.id = ex_sub.id
+                self.update_subscription(subscription)
+                return ex_sub.id
+
         url = urljoin(self.base_url, 'v2/subscriptions')
         headers = self.headers.copy()
         headers.update({'Content-Type': 'application/json'})
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -23,7 +23,7 @@ copyright = '2021, RWTH Aachen University, ' \
 author = 'EON EBC'
 
 # The full version, including alpha/beta/rc tags
-release = '0.1.5'
+release = '0.1.6'
 
 # The short X.Y version.
 version = '0.1'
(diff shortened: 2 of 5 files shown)


filip/clients/ngsi_v2/cb.py
```diff
@@ -880,18 +881,39 @@ def get_subscription_list(self,
...
-            object: 
+            str: Id of the (created) subscription
 
         """
+        existing_subscriptions = self.get_subscription_list()
+
+        sub_hash = str(subscription.dict(include={'subject', 'notification'}))
+        for ex_sub in existing_subscriptions:
+            if sub_hash ==\
+                    str(ex_sub.dict(include={'subject', 'notification'})):
```
tstorek, line 906: If you cast it to a str, why don't you use to_json(include:{})?

filip/clients/ngsi_v2/cb.py
```diff
@@ -880,18 +881,39 @@ def get_subscription_list(self,
...
+                self.logger.info("Subscription already exists")
+                if update:
```
tstorek, line 908: This only checks for existence of update meaning that any variable but "0, False, None" will trigger this

filip/clients/ngsi_v2/cb.py
```diff
@@ -880,18 +881,39 @@ def get_subscription_list(self,
             self.log_error(err=err, msg=msg)
             raise
 
-    def post_subscription(self, subscription: Subscription) -> str:
+    def post_subscription(self, subscription: Subscription,
+                          update: Optional[bool] = False) -> str:
```
tstorek, line 885: You don't need Optional here because you have the default value. Since you definetly check for update it is  not optional.

filip/clients/ngsi_v2/cb.py
```diff
@@ -880,18 +881,39 @@ def get_subscription_list(self,
...
         """
         Creates a new subscription. The subscription is represented by a
         Subscription object defined in filip.cb.models.
 
-        Args:
-            subscription:
+        If the subscription already exists, the adding is prevented and the id
+        of the existing subscription is returned.
```
tstorek, line 891: Please elaborate how subscriptions are compared to each other. This is not trivial!

tstorek (changes requested):

filip/clients/ngsi_v2/cb.py
RedingJeff: Yes you are right, I wanted to use Optional as a Hint for the user, but that is not how it is correctly used.

filip/clients/ngsi_v2/cb.py
RedingJeff: How would you recommend checking the condition? As update is of type bool and has the default value False, this should be the correct logic. The user can technically enter a string as update parameter, but he will get a type notification. 

I can change it to update==True, to only allow True as correct value, but I think that derives from the standard code style guidelines. What do you think?

filip/clients/ngsi_v2/cb.py
RedingJeff: I needed to cast it to str as he would otherwise compare the object pointers. I did not found a better suited methode to achieve this result. I can not use to_json() as subscription is a pydantic model and not a pandas dataframe.

filip/clients/ngsi_v2/cb.py
tstorek: This does fit now :)

tstorek:
closes #22

Issue #24, opened 2021-09-14
Add MQTT notifications to CB_client


tstorek:
**Is your feature request related to a problem? Please describe.**
With the new version of the Orion Context Broker notifications via MQTT are supported.
https://fiware-orion.readthedocs.io/en/3.2.0/user/mqtt_notifications/index.html

**Describe the solution you'd like**
We want to add this functionality in our notifications models.
Furthermore, we need to introduce the deprecated query parameter for skipping the initial notification for backward compatibility. If set the the client will check the version of the context broker first and then return a warning to the user.

**Describe alternatives you've considered**
None

tstorek:
closed by #25

PR #25, opened 2021-09-14
updated example for iot  over mqtt

tstorek:
For #24

### DIFF
--- a/examples/iot_mqtt_example.py
+++ b/examples/iot_mqtt_example.py
@@ -28,6 +28,7 @@ logging.basicConfig(
 logger = logging.getLogger('filip-iot-example')
 
 # Before running the example you should set some global variables
+# Please, enter your URLs here!
 CB_URL = "http://yourHost:yourPort"
 IOTA_URL = "http://yourHost:yourPort"
 MQTT_BROKER_URL = "http://yourHost:yourPort"
@@ -35,6 +36,7 @@ DEVICE_APIKEY = 'filip-iot-example-device'
 SERVICE_GROUP_APIKEY= 'filip-iot-example-service-group'
 FIWARE_SERVICE = 'filip'
 FIWARE_SERVICE_PATH = '/iot_examples'
+
 
 if __name__ == '__main__':
     # Since we want to use the multi-tenancy concept of fiware we always start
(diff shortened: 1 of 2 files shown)


filip/models/ngsi_v2/registrations.py
```diff
@@ -0,0 +1,148 @@
...
+    QUERY = "query", "This provider only supports request forwarding to " \
+                     "query data."
+    UPDATE = "update", "This provider only supports request forwarding to " \
+                       "update data."
+    ALL = "all", "This provider supports both query and update forwarding " \
+                 "requests. (Default value)"
+
+
+class Provider(BaseModel):
+    http: AnyHttpUrl = Field(
```
RedingJeff, line 35: In Issue #29 I was currently fixing the issue here. 
Provider does need an **url** field inside the **http** field containing the given httpadresse. Same as Notification in the subscription class, where the HTTP class was used. 

Maybe the HTTP class should be on a more global scale so that the Provider can here also use it

filip/models/ngsi_v2/registrations.py
```diff
@@ -0,0 +1,148 @@
...
+        description="It is used to convey parameters for providers that "
```
RedingJeff, line 36: In Issue #29 I was currently fixing the issue here.
Provider does need an url field inside the http field containing the given httpadresse. Same as Notification in the subscription class, where the HTTP class was used.

Maybe the HTTP class should be on a more global scale so that the Provider can here also use it.

As the file was drastically changed in location and name, I suggest that it is for the best if we discard the changes made in Issue #29 and you simply add these changes here. A small test for registrations would also be very useful (I just started with it in Issue 29)

RedingJeff (changes requested):

tstorek:
@RedingJeff i figured that there are additional models reused. We can either move back to a single file "context.py" and make the naming a little more descriptiv or "context_base.py" containing shared models. Perhaps the single file wil be the easiest. Inheritance would surrogate an artificial hierarchy that actually not exist.

Any thoughts on this?

RedingJeff:
@tstorek I think it is definitly better to have them in seperate files, it makes the structure of those the seperated concepts much cleaner. Maybe we can prevent the issue with the suggested hierarchy, by not naming the file containing the shared models "context_base", but "global_models", "shared_models" or something like that ?

tstorek:
@RedingJeff Alright. Then i will simply introduce models.ngsi_v2.context.base.py or models.ngsi_v2.cb.base.py that will collect all shared for the ocb.

Hence:
models.ngsi_v2.cb.base.py
models.ngsi_v2.cb.context.py
models.ngsi_v2.cb.subscriptions.py
models.ngsi_v2.cb.registrations.py

PR #28, opened 2021-09-15
feat: Issue#27 deleting

RedingJeff:
The wanted functionality is implemented and tested with a test_method.

One entity can be assigned to multiple devices, I think this is more of an edge case, but naturally still needs to be treated in the program logic. 
I was unsure if in that case the delete_device(delete_entity=True), the user should be warned that the entity could not be deleted as it is belongs to more than this device, or if an exception should be called. At the moment an exception is called.


tstorek:
@RedingJeff could you please resolve the conflicts before I will review?

tstorek (approved):

PR #31, opened 2021-09-22
Add "autoprovision" to the class ServiceGroup

orange95:
Add "autoprovision" to the class ServiceGroup


tstorek (changes requested):
Thanks, for the change. As I asked in the issue please add proper documentation. is there any default?
docs are found here:
https://iotagent-node-lib.readthedocs.io/en/latest/api/index.html#service-group-api

"optional boolean: If false, autoprovisioned devices (i.e. devices that are not created with an explicit provision operation but when the first measure arrives) are not allowed in this group. Default (in the case of omitting the field) is true."

PR #36, opened 2021-09-24
Issue#34 Teardowns to remove subscriptions

RedingJeff:
Added proper tear-downs to remove the subscriptions, to ensure that even in case of a test failure the state is cleaned. 

Could not resolve the failing tests in the CI/CD pipeline, locally the always work.

It could be observed that the tests do not fail always in the pipeline the subscription test in test_ngsi_v2_context.py fails rarley and if it shows a very strange behavior of missing one optional data field in the subscription entity.

The subscription test in test_ngsi_v2_client.py fails more often, currently no logical reason can be given

### DIFF
--- a/tests/clients/test_ngsi_v2_cb.py
+++ b/tests/clients/test_ngsi_v2_cb.py
@@ -333,10 +333,7 @@ class TestContextBroker(unittest.TestCase):
             id3 = client.post_subscription(sub2)
             self.assertNotEqual(id1, id3)
 
-            # Clean up
-            subs = client.get_subscription_list()
-            for sub in subs:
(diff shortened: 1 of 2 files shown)


tstorek:
@RedingJeff maybe we can test by introducing a small latency that gives fiware a little time to settle the subscriptions. This may solve the issue. I think that this is the only difference between the two scenarios but the operating system.

RedingJeff:
@tstorek I had already introduced a 1 second sleep between the commands to prevent a system latency from interfering with the test. But maybe it needs to be larger, I will try it out. If this would really solve the issue, we need to incorporate a waiting logic into the method itself.

PR #48, opened 2021-10-09
feat: Issue#41 introduce concurrent testing

### LINKED ISSUE #41: Introduce concurrent testing
Currently, concurrent testing will automatically interfere with each other because the `servicepath`s are static. Therefore, the tests will all use the same tenants which will result in failing tests, if they try to access the same resources.

The idea is to introduce a `TestingSettings`-class that will check for a local environment for its variables. If a `*.env.` is not available it will try to use environment variables. These variables should be set in the CI because the CI should not contain a file with information about the used servers. The `servicepath` should then be either autogenerated or set in the first or second way.

tstorek:
closes #41


filip/utils/cleanup.py
```diff
@@ -0,0 +1,157 @@
...
+        None
+    """
+    # create client
+    client = IoTAClient(url=url, fiware_header=fiware_header)
+
+    # clear groups
+    for group in client.get_group_list():
+        client.delete_group(resource=group.resource,
+                            apikey=group.apikey)
+
```
RedingJeff, line 68: Is a device always part of a group?

filip/utils/cleanup.py
```diff
@@ -0,0 +1,157 @@
...
+    # clear data
+    try:
+        for entity in client.get_entities():
+            client.delete_entity(entity_id=entity.entityId,
+                                 entity_type=entity.entityType)
+    except RequestException as err:
+        if err.response.status_code == 404:
+            try:
+                err.response.json()['error'] == 'Not Found'
+            except KeyError:
```
RedingJeff, line 96: Shouldn't be also an error raised if the 'error' entry is not 'not found' ?

tests/clients/test_ngsi_v2_timeseries.py
```diff
@@ -31,120 +32,165 @@ def setUp(self) -> None:
...
-        except:
-            pass
-
-        self.client.close()
+        clear_all(fiware_header=self.fiware_header,
+                  cb_url=settings.CB_URL,
+                  ql_url=settings.QL_URL)
+
+        self.ql_client.close()
         self.cb_client.close()
```
RedingJeff, line 196: Is this as well as each other teardown still needed? If the test fails is the header clean_up then not executed? And even if, each test cleans the state before its execution. 
In fact it probably does not harm us if we also execute a clean in the teardown, but I think we could skip it?

tests/models/test_base.py
```diff
@@ -41,29 +47,28 @@ def test_fiware_header(self):
...
-                try:
-                    client.fiware_service_path = path
-                    entities = client.get_entity_list()
-                    client.update(entities=entities, action_type='delete')
-                except RequestException:
-                    pass
+                header = FiwareHeader(
+                    service=settings.FIWARE_SERVICE,
+                    service_path=path)
+                clear_all(fiware_header=header, cb_url=settings.CB_URL)
```
RedingJeff, line 73: Ok here is a good use for the teardown methode, as it cleans the generated paths.

filip/utils/cleanup.py
tstorek: @RedingJeff actually, with clear we want that error message because it tells us that not entries are in the database

RedingJeff: @tstorek OK, but does the == statement here have any effect? Doesnt the code only tries if it can access the "error" key of the json() without error?

tests/clients/test_ngsi_v2_timeseries.py
tstorek: @RedingJeff currently, if the test fails an error is raised. Hence, everything from there will not be executed. I can check if we can catch and reraise, but I think that is what teardown is for. It is basically a backup.

filip/utils/cleanup.py
tstorek: @RedingJeff good point 👍

tests/clients/test_ngsi_v2_timeseries.py
RedingJeff: Then we should simply leave it like this. It has no negative effect, better to have a backup than code that is 0.5 seconds faster in testing, but fails due to state issues.

filip/utils/cleanup.py
RedingJeff: Nevermind, i was confused by the comment "#clear registrations", what should probably be "#clear devices". Therefore I thought we only clear the devices here indirectly over their groups.

PR #49, opened 2021-10-10
fix: 38 substitute fuzzywuzzy with rapidfuzz

### LINKED ISSUE #38: substitute fuzzywuzzy with rapidfuzz
Due to Licence issues, we should substitute the fuzzywuzzy package used for unit validation with RapidFuzz.

https://github.com/maxbachmann/RapidFuzz

The usage of RapidFuzz should work in very similar way.

tstorek:
closes  #38


tstorek:
can be merged after #48

PR #61, opened 2021-10-18
feat: 60 static device attribute metadata type

RedingJeff:
The metadata logic of the device attributes is now the same as the logic for the context attributes. 

It is currently code duplication needed (metadata validator, nealry all device attr fileds). What we could do to prevent it -> make BaseAttribute (of iot.py) inherrit from ContextAttribute (context.py). 

The only difference between these two fields are:

- BaseAttribute, has more fields (no problem)
- ContextAttribute has the field "value", and BaseAttribute not (that is a problem).

This solution would merge the currently seperated context and iot worlds, but remove code duplications. Build the following inheritance structure:

- BaseAttribute (new, equals current ContextAttribute without value field)
- ValueAttribute (only has value field)

=> 
- ContextField(BaseAttribute, ValueAttribute)
    - all current children remain unchanged
- IoTaBaseAttribute(BaseAttribute)
    -  DeviceAttribute(IoTaBaseAttribute)
    -  LazyDeviceAttribute(IoTaBaseAttribute)
    -  StaticDeviceAttribute(IoTaBaseAttribute, ValueAttribute)
- DeviceCommand(Basemodel) (unchanged)


RedingJeff:
I also added the default value None for each model field that had the type Optional but no default value.

Now their is no hint error if the user does not specifies that an optional field gets assigned the value None. Resulting in a clearer user experience

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -34,7 +36,7 @@
     DeviceCommand, \
     DeviceAttribute, \
     ServiceGroup, \
-    StaticDeviceAttribute
+    StaticDeviceAttribute, TransportProtocol
```
tstorek, line 39: please do long imports in list in vertical order

tests/models/test_ngsi_v2_iot.py
```diff
@@ -5,8 +5,8 @@
 import unittest
 from typing import List
 
-from filip.models.ngsi_v2.iot import BaseAttribute, DeviceCommand, ServiceGroup, \
-    Device, TransportProtocol
+from filip.models.ngsi_v2.iot import DeviceCommand, ServiceGroup, \
+    Device, TransportProtocol, IoTaBaseAttribute
```
tstorek, line 9: IoTa schould be IoTA. this would correlate to the other class names

tests/models/test_ngsi_v2_iot.py
```diff
@@ -69,28 +69,28 @@ def test_fiware_safe_fields(self):
...
+                              entity_name=string, entity_type=string)
             DeviceCommand(name=string, type="name")
             ServiceGroup(entity_type=string, resource="", apikey="")
             Device(device_id="", entity_name=string, entity_type=string,
                    transport=TransportProtocol.HTTP)
 
         # Test for the special-string protected field if all strings are blocked
         for string in special_strings:
-            self.assertRaises(ValidationError, BaseAttribute,
+            self.assertRaises(ValidationError, IoTaBaseAttribute,
```
tstorek, line 81: usually when you do this using the context protocol is mcuh cleaner.

```python
with self.assertRaises(ValidationError):
    by failing function
```

tstorek: but i think taht your approach is also valid

tstorek (changes requested):
hi, thanks for the pull. only minor remarks

tstorek (approved):

PR #70, opened 2021-11-03
fix: 69 test clean up uses wrong method to clean entities

RedingJeff:
Test deletion fixed.

I saw that the use of delete_entity for all entity was much slower than using the batch methode (16s to 6s).

Therefore I implented a methode that can also be used by endusers to efficiently delete entities as batch.
The methode calls update(delete) twice (with controlled logic)

Additionally I removed a few clean-all() class that were redundant, as each test calls it in the tear_down

### DIFF
--- a/filip/clients/ngsi_v2/cb.py
+++ b/filip/clients/ngsi_v2/cb.py
@@ -547,6 +547,39 @@ class ContextBrokerClient(BaseHttpClient):
             self.log_error(err=err, msg=msg)
             raise
 
+    def delete_entities(self, entities: List[ContextEntity]) -> None:
+        """
+        Remove a list of entities from the context broker. This methode is
+        more efficient than to call delete_entity() for each entity
+
+        Args:
+            entities: List[ContextEntity]: List of entities to be deleted
+        Raises:
+            Exception, if
+        Returns:
+            None
+        """
+
+        # update() delete, deletes all entities without attributes completely,
+        # and removes the attributes for the other
+        # The entities are sorted based on the fact if they have
+        # attributes.
+        entities_with_attributes: List[ContextEntity] = []
+        for entity in entities:
+            property_names = [key for key in entity.dict() if key not in
+                              ContextEntity.__fields__]
+            if len(property_names) > 0:
+                entities_with_attributes.append(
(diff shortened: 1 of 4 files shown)


RedingJeff:
The test if an entity has attributes is currently not the cleanest, but the issue30 introduces a better entity attribute model interface.

RedingJeff:
The test in the CI currently fails, this is already fixed in Pullrequest 66.

filip/clients/ngsi_v2/cb.py
```diff
@@ -547,6 +547,39 @@ def delete_entity(self, entity_id: str, entity_type: str = None) -> None:
...
 
+    def delete_entities(self, entities: List[ContextEntity]) -> None:
+        """
+        Remove a list of entities from the context broker. This methode is
+        more efficient than to call delete_entity() for each entity
+
+        Args:
+            entities: List[ContextEntity]: List of entities to be deleted
+        Raises:
+            Exception, if
```
tstorek, line 558: Could you please add the rest of raises section?

filip/utils/cleanup.py
```diff
@@ -5,6 +5,8 @@
 
 @author Thomas Storek
 """
+import time
+
```
tstorek, line 9: remove empty line

requirements.txt
```diff
@@ -11,5 +11,5 @@ paho-mqtt>=1.5.1
 pandas_datapackage_reader>=0.18.0
 tables>=3.6.1
 python-Levenshtein>=0.12.2
-paho-mqtt
+paho-mqtt==1.5.1
```
tstorek, line 14: why not >= 1.5.1?

tstorek (changes requested):
thanks. only minor remarks

requirements.txt
RedingJeff: Paho introduced breaking changes in version 1.6.0 making the code fail. As I did not find a good changelog, I decided that it should be ok to keep the older version, as it is is only used in the test case. 
If wanted it could be made working with the newer paho version, with a bit of effort

tstorek (approved):

PR #71, opened 2021-11-04
feat: 14 adding mqtt client

tstorek:
I had to skip some minor tests. due to other issues. Setting the timestamp does currently not work due to datetime format parsing.

The command acknowledgment does not work. I could not fix this yet but this will be an additional issue.


RedingJeff (approved):
Looks good

PR #73, opened 2021-11-12
feat: 47 quantum leap request pagination

tstorek:


filip/clients/ngsi_v2/quantumleap.py
```diff
@@ -515,21 +610,27 @@ def get_entity_values_by_id(self,
...
-                                   from_date=from_date,
-                                   to_date=to_date,
-                                   last_n=last_n,
-                                   limit=limit,
-                                   offset=offset,
-                                   georel=georel,
-                                   geometry=geometry,
-                                   coords=coords)
-        return TimeSeries(entityId=entity_id, **res)
+        res_q = self.__query_builder(url=url,
```
RedingJeff, line 613: There seems to multiple code duplications in the methods get_entity_values_by_id(613), get_entity_attr_by_id(700), get_entity_attr_values_by_id(761), get_entity_by_type(816), get_entity_values_by_type(870), et_entity_attr_by_type(950), get_entity_attr_values_by_type(1053).

Without a more detailed understanding of the used TimeSeries Model is it hard to say how easy parts of these redundancies can be eliminated with helper methods.  I suggest taking a look if there exists a good possibility to remove a bit of the duplication else keep it as is.

RedingJeff (changes requested):

filip/clients/ngsi_v2/quantumleap.py
tstorek: @RedingJeff  if you look closer you will recognize that all functions call the same query builder. sure this has to be the same. However, parsing always differs from function to function

RedingJeff: @tstorek okay. When looking from above their structure was quite similar, but as said it is hard to know without knowing the structure behind in detail how much of the current similarities could be grouped together.

If the parsing differs that much then I have no more comments on this pull request. It can be merged

RedingJeff (approved):

Issue #30, opened 2021-09-20
Ontologies to Semantic 


RedingJeff:
**Describe the solution you'd like**
Take the semantic logic that is currently situated in the SENSE fronted and transfer it over into Filip. 
Further this logic needs to be transformed to better synergies with Filip entities. 

Instead of the currently used instances that are only transformed to Filip entities when saving the state, a wrapper will be written to simply extend the entities with the needed semantic functionalities.

To make the storage and reparsing of the semantic vocabulary much easier the ontology files will be saved inside the pydantic vocabulary object, which can be stored as json.

If this feature is finished a user can provide a semantic structure to Filip by giving ontologies, and use this structure to create models, connect them and model a state.

PR #75, opened 2021-11-18
Feat: Patch Methods

RedingJeff:
I added the Patch device method and the tests as described in issue 74.
I discovered that Patch Entity and the context model changes were not yet live on development and only local in #30. 

I therefore bundled here all these changes to be reviewed together.

The made changes can be summarized as:
- Patch Methods (Device and Entity)
- ContextEntity Remodel (to make attribute interface more clear to the user)
- Helper functions (does entity/device exists ?)
- Test functions to test new and a few old uncovered methods
- Change of comments to make functionality more clear

### DIFF
--- a/filip/clients/ngsi_v2/cb.py
+++ b/filip/clients/ngsi_v2/cb.py
@@ -1408,9 +1408,8 @@ class ContextBrokerClient(BaseHttpClient):
         """
         try:
             self.get_entity(entity_id=entity_id, entity_type=entity_type)
-        except requests.RequestException as ex:
-            # couldn't find a better way to extract error code
-            if not str(ex)[0:36] == "404 Client Error: Not Found for url:":
+        except requests.RequestException as err:
+            if not err.response.status_code == 404:
                 raise
             return False
         
@@ -1420,14 +1419,14 @@ class ContextBrokerClient(BaseHttpClient):
                      entity: ContextEntity,
                      old_entity: Optional[ContextEntity] = None) -> None:
         """
-           Takes a given entity and updates the state in the CB to match it.
-           Args:
-               entity: Entity to update
-               old_entity: OPTIONAL, if given only the differences between the
-                           old_entity and entity are updated in the CB.
-                           Other changes made to the entity in CB, can be kept.
-           Returns:
-               None
+        Takes a given entity and updates the state in the CB to match it.
+        Args:
+           entity: Entity to update
+           old_entity: OPTIONAL, if given only the differences between the
+                       old_entity and entity are updated in the CB.
+                       Other changes made to the entity in CB, can be kept.
+        Returns:
+           None
         """
 
         new_entity = entity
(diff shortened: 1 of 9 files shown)


filip/clients/ngsi_v2/cb.py
```diff
@@ -523,7 +524,8 @@ def update_entity(self,
             raise
 
     def delete_entity(self, entity_id: str, entity_type: str,
-                      delete_devices: bool = False) -> None:
+                      delete_devices: bool = False,
+                      iota_url: AnyHttpUrl = settings.IOTA_URL) -> None:
```
tstorek, line 529: Well, you may also have entities that do not correlate to a device in the IoTA agent. Hence document it that it is required when delete_devices is True

filip/clients/ngsi_v2/cb.py
```diff
@@ -1386,6 +1390,137 @@ def post_command(self,
...
+        Returns:
+            bool; True if entity exists
+
+        Raises:
+            RequestException, if any error occurres (e.g: No Connection),
+            except that the entity is not found
+        """
+        try:
+            self.get_entity(entity_id=entity_id, entity_type=entity_type)
+        except requests.RequestException as ex:
```
tstorek, line 1411: it should be err instead of ex

filip/clients/ngsi_v2/cb.py
```diff
@@ -1386,6 +1390,137 @@ def post_command(self,
...
+            # couldn't find a better way to extract error code
+            if not str(ex)[0:36] == "404 Client Error: Not Found for url:":
```
tstorek, line 1413: the requests package returns an objects that contains the complete server response object --> err.response.status_code ==404 you can also access all other HTTP response parameters there.

filip/clients/ngsi_v2/cb.py
```diff
@@ -1386,6 +1390,137 @@ def post_command(self,
...
+            return False
+        
+        return True
+
+    def patch_entity(self,
+                     entity: ContextEntity,
+                     old_entity: Optional[ContextEntity] = None) -> None:
+        """
+           Takes a given entity and updates the state in the CB to match it.
+           Args:
```
tstorek, line 1424: Empty line is missing

filip/clients/ngsi_v2/cb.py
```diff
@@ -1386,6 +1390,137 @@ def post_command(self,
...
+                raise
+            return False
+        
+        return True
+
+    def patch_entity(self,
+                     entity: ContextEntity,
+                     old_entity: Optional[ContextEntity] = None) -> None:
+        """
+           Takes a given entity and updates the state in the CB to match it.
```
tstorek, line 1423: overindentation

filip/clients/ngsi_v2/cb.py
```diff
@@ -1386,6 +1390,137 @@ def post_command(self,
...
+        # Check the differences between the attributes of old and new entity
+        # Delete the removed attributes, create the news,
+        # and update the existing if necessary
+        old_attributes = old_entity.get_properties()
+        old_attributes.extend(old_entity.get_relationships())
+
+        new_attributes = new_entity.get_properties()
+        new_attributes.extend(new_entity.get_relationships())
+
+        # Manage attributes that existed before
```
tstorek, line 1480: actually this applies to all attributes that are provided via registration. Think that you finally have to check if there are corresponding registrations. https://fiware-tutorials.readthedocs.io/en/latest/context-providers.html check for the registrations and check them if they match the entity attributes. All matching attributes do probably come from registrations

filip/clients/ngsi_v2/iota.py
```diff
@@ -461,6 +477,131 @@ def delete_device(self, *, device_id: str,
...
+        if patch_entity:
+            from filip.clients.ngsi_v2 import ContextBrokerClient
+            with ContextBrokerClient(
+                    url=cb_url,
+                    fiware_header=self.fiware_headers) as client:
+
+                client.patch_entity(
+                    entity=build_context_entity_from_device(device))
+
+    def does_device_exists(self, device_id: str) -> bool:
```
tstorek, line 591: please correlate with the context broker implementation. Does it actually need to be a public method? All other users could simple use get_device and check the response

filip/models/ngsi_v2/context.py
```diff
@@ -203,49 +203,173 @@ def _validate_properties(cls, data: Dict):
...
+        Returns:
+            NamedContextAttribute
+
+        """
+        for attr in self.get_attributes():
+            if attr.name == attribute_name:
+                return attr
+        raise KeyError
+
+    def get_properties(
```
tstorek, line 354: why not also a function for adding a property? to similar to add_attribute which could also add a relationship?

tests/clients/test_ngsi_v2_iota.py
```diff
@@ -177,15 +178,16 @@ def test_deletions(self):
                         entity_name=entity_id,
                         entity_type='Thing2',
                         protocol='IoTA-JSON',
-                        transport='MQTT',
+                        transport='HTTP',
```
tstorek, line 181: doesn't this make other tests fail? you are not using any HTTP here, right?

tstorek (changes requested):
Thanks, found some minor stuff

tests/clients/test_ngsi_v2_iota.py
RedingJeff: This object is only used in this specific test, it is locally created. I changed it to test out if it was the source of an issue that I was having, but that was not the case.
For this test it has no influence which transport was used, I just simply did not change it back, as it is not important.

filip/models/ngsi_v2/context.py
RedingJeff: For the naming I oriented myself on the existing ideas. Each (Named)ContextAttribute is an attribute, if it further has the type RELATIONSHIP it is also a relationship else a property. 
As it is not important for the adding process if you add a property or relationship all 'add' functions are kept in "superclass style" (attribute), same with update and delete. 

Only the getting functions are interesting for the user to directly get Relationships or all properties (as it was implemented before), therefore only at that place the additional filtering functions exist.

filip/clients/ngsi_v2/iota.py
RedingJeff: I added it exactly for all other users, as in my opinion it is a good addition to method space. As I see it it has two advantages:

1. The user directly sees that the functionality exists, and does not work itself into it to discover that get_device throws an error if it does not exist.
2. It makes the user code cleaner, instead off always needing to use a try catch block.

The ContextBroker got a correlating method : does_entity_exists(..).
I could change it in my code to use try and catch's directly in the methods, but with the reasons above I would argument to keep these two methods.

filip/clients/ngsi_v2/cb.py
RedingJeff: In which line the empty line should be inserted?, I do not find a violation of the style guide  (Except the now fixed over-indent)

tstorek:
please adjust the changelog

filip/clients/ngsi_v2/cb.py
RedingJeff: As the method does not fail if no correlating device(s) exists, is it necessary to make that annotation?
The current statement holds True in all case: Delete any device that corresponds to this entity.

tstorek (approved):

PR #78, opened 2021-11-25
Support for python 3 9

RedingJeff:
I changed the dataformat from HDF to CSV and therefore could drop the requirement for tables.

The tests are all passing, and I confirmed that I can now locally execute Filip with standard Python 3.9


tstorek (approved):
thank you!

PR #96, opened 2021-12-06
feat: Semantic Model Search

RedingJeff:
As requested I implemented a search function for the semantic classes/models.

A test was added as well as an entry into the examples.

### DIFF
--- a/examples/semantics/semantics_model_example.py
+++ b/examples/semantics/semantics_model_example.py
@@ -65,7 +65,7 @@ if __name__ == '__main__':
     #
     # More details about the uniqueness later
 
-    # ## 2.1 Instantiating a class
+    # ## 2.1. Instantiating a class
     #
     # We can create an instance of a class by simply instantiating and
     # passing the id and header as parameters.
@@ -87,7 +87,7 @@ if __name__ == '__main__':
     semantic_manager.set_default_header(header)
     my_building = Building(id="building1")
 
-    # ## 2.2 Immutability
+    # ## 2.2. Immutability
     #
     # These defining information of an instance are immutable and can not
     # be changed after the creation.
@@ -100,7 +100,7 @@ if __name__ == '__main__':
     print(my_building.get_identifier())
     print("")
 
-    # ## 2.3 Device Classes
+    # ## 2.3. Device Classes
     #
     # Device-classes are a subtype of normal classes, they work as normal
     # classes, but posses additional field types and information about the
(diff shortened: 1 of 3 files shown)


examples/semantics/semantics_model_example.py
```diff
@@ -464,6 +479,8 @@
     p1.device_settings.endpoint = "http://test2.com"
     p1.device_settings.transport = TransportProtocol.AMQP
```
tstorek, line 480: why not MQTT?

examples/semantics/semantics_model_example.py
```diff
@@ -122,6 +122,22 @@
...
     my_sensor.device_settings.transport = TransportProtocol.HTTP
 
+    # # 2.4. Finding a fitting class
+    #
+    # If you are unsure which semantic class could be the right one to model
+    # your current need, you can search for a fitting possibility:
+
+    fitting_names = semantic_manager.find_fitting_model("SmokeSensor")
+
+    print("\u0332".join("Found models for 'SmokeSensor':"))
```
tstorek, line 132: Is the join really necessary? I think that in an example it is not taht important and it disturbs a little the readability

examples/semantics/semantics_model_example.py
```diff
@@ -65,7 +65,7 @@
     #
     # More details about the uniqueness later
 
-    # ## 2.1 Instantiating a class
+    # ## 2.1. Instantiating a class
```
tstorek, line 68: why do you put dots here? that is rather unusual when you structure a text into subsections --> have a lokk at latex

filip/semantics/semantics_manager.py
```diff
@@ -992,3 +992,26 @@ def _get_added_and_removed_values(
                 if old_settings[key] is not current_settings[key]:
                     new_settings[key] = current_settings[key]
                 instance.device_settings.__setattr__(key, new_settings[key])
+
+    def find_fitting_model(self, search_term: str) -> List[str]:
```
tstorek, line 996: you may allow a limit arg

tstorek (changes requested):
looks good. only minor things :)

examples/semantics/semantics_model_example.py
RedingJeff: With the join it is in the same style as the rest of the example. I chose this print, so that the output that contains a lot of information is more readable, as it underlines the explanation prints()

tstorek (changes requested):
please add hte changelog :)

tstorek (approved):

PR #97, opened 2021-12-06
Fix: Failing test (command_callback)

RedingJeff:
In the past multiple tests failed to the command_callback test for ultralight. 

I now added a wait an retry loop in the test to reduce the chance of a test failure.

While it can not be 100% guaranteed that the tests will now never fail, none of the multiple tries failed, so at least the chance of failure seems to be reduced.

While testing the same issue was discovered for the JSON_command_callback, so the same fix was applied.


RedingJeff:
The last merge test failed ^^ the loop seems to be no solution.
It seems the test fails before the state is looked up. Maybe rerun the whole test on failure?

tstorek:
agreed to accept that the test sometimes fails. This is usually due to server configuration issues. Try to rerun the test_stage

PR #100, opened 2021-12-08
Fix: Fixed header loading

RedingJeff:
Fixed the mentioned issue.

Also added a quick fix for two tests that were failing depending on the runtime environment (as the cb-client used the environment url)

### DIFF
--- a/examples/semantics/semantics_model_example.py
+++ b/examples/semantics/semantics_model_example.py
@@ -516,6 +516,80 @@ if __name__ == '__main__':
 
     # # 5. Visualisation
     #
-    # Concluding we can always have a look inside the local loaded state by
-    # calling:
+    # ## 5.1 Simple Debug Visualisation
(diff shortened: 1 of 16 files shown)


RedingJeff:
- Due to the work environment, this pull request holds a set of different features and fixes:

- Improved the parser and postprocessor: 
-        Now able to merge classes additions given by an additional ontology ( a class can now have multiple sources)
-        Multiple Fixes
- Provided more interface methods to interact with the semantic state
-        public methode to precheck if a name is valid for a DeviceAtrribute
-        public state validation methode, to be able to check semantic state, before saving
- Prevented an invalid ID to be given for a semantic instance
- Cytoscape computation methode (get the computed data, to display in a cytoscape plugin)
- Fixed: Does_entity_exists (depending on the call the handler tried to dereference a None Pointer)

PR #114, opened 2022-01-21
fix: 113 missing attribute requests in cb

tstorek:


tstorek:
@RedingJeff Apparently, there are patch functions in the specs that we simply overlooked. could your maybe check if the functions might break things?

RedingJeff:
@tstorek Do you have a link to the documentation of these functions?

tstorek:
@RedingJeff https://fiware.github.io/specifications/ngsiv2/stable/

Please check the entities section

filip/clients/ngsi_v2/cb.py
```diff
@@ -601,32 +592,159 @@ def delete_entities(self, entities: List[ContextEntity]) -> None:
...
 
         Args:
-            entity (ContextEntity):
-            append (bool):
-            options:
+            entity_id: Entity id to be updated
+            entity_type: Entity type, to avoid ambiguity in case there are
+                several entities with the same entity id.
+            attrs: List of attributes to update or to append
+            append: If `False` the entity attributes are updated (if they
```
RedingJeff, line 617: I find the "append" flag quite confusing. 
In summary: if true non-existing are appended, and also if false they are also appended. 
The only difference is with updating ?

RedingJeff (approved):
Tutorial look good.

As far as i see it from the update methode documentation, they definitly do not replace the patch methode. 
The patch methode also handles class renames etc and most importantly for the semantic plugin it considers that a Fiware entry could be manipulated by mutliple users/servies at the same time and only tries to push the made changes to Fiware thus merging as best as possible the different states, and not only overwritting the live state.

examples/e5_context_broker_subscriptions_mqtt.py
```diff
@@ -0,0 +1,187 @@
...
+    sub = Subscription(**sub_example)
+
+    # Posting an example subscription for Room1. Make sure that you store the
+    # returned id because you might need for later updates of the subscription.
+    sub_id = cb_client.post_subscription(subscription=sub)
+
+    # # 3 setup callbacks and the MQTT client
+    #
+    # define callbacks for the mqtt client. They will be triggered by
+    # different events. Do not change there signature!
```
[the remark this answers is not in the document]
tstorek: done!

filip/clients/ngsi_v2/cb.py
tstorek: i noticed that but simply used it from the specs. I changed it to `append_strict`.

PR #138, opened 2022-04-04
feat: compare subscriptions for different dict formats

### LINKED ISSUE #137: Check if subscription already exists
I would like to avoid posting duplicate subscriptions. 
I've already written a working function for this:

```python
# Get existing subs
subscription_list = cb_client.get_subscription_list()
# Build a new one
subscription = Subscription(**my_data)
# Only post if it not exists already:
if not subscription_already_exists(subscription, existing_subscriptions=subscription_list):
    logger.info("Posting new subscription: %s", subscription.json(indent=2))
    cb_client.post_subscription(subscription=subscription)
```

The function `subscription_already_exists`:

```python
def subscription_already_exists(subscription, existing_subscriptions):
    sub = subscription.dict(exclude={"id"}).copy()
    for _existing_sub in existing_subscriptions:
        ex_sub = _existing_sub.dict(exclude={"id"}).copy()
        equals = sub_dict_is_equal(ex_sub, sub)
        if equals:
            return True
    return False  # If still here, it's not in the list


def sub_dict_is_equal(first, second):
    if first.keys() != second.keys():
        logger.warning(
            "Subscriptions contain a different set of fields. "
            "Only comparing to new fields of the new one."
        )
    for k, v in first.items():
        ex_value = second.get(k, None)
        if isinstance(v, dict):
            equal = sub_dict_is_equal(v, ex_value)
            if equal:
                continue
            else:
                return False
        if not v:
            ex_value = bool(ex_value)
            v = bool(v)
        if v != ex_value:
            return False
    return True

```

The nasty recursive comparison is due to filip and fiware interaction. Empty attributes in the notification dict are sometimes None and sometimes `[]`. Without this, the feature would be even shorter.

I would opt to move the function to `post_subcription` and add a kwarg `check_duplicate` or some better fitting name.

FWuellhorst:
Closes #137 .

@tstorek : I added the feature. I will write tests etc. once you approve the basic idea.


filip/clients/ngsi_v2/cb.py
```diff
@@ -1717,6 +1720,54 @@ def patch_entity(self,
...
+            )
+        for k, v in first.items():
+            ex_value = second.get(k, None)
+            if isinstance(v, dict) and isinstance(ex_value, dict):
+                equal = self._subscription_dicts_are_equal(v, ex_value)
+                if equal:
+                    continue
+                else:
+                    return False
+            if not _value_is_not_none(v) and not _value_is_not_none(ex_value):
```
djs0109, line 1764: Should it be "if not _value_is_not_none(v) **or** not _value_is_not_none(ex_value)" here? Otherwise the none existing keys will not be ignored.

filip/clients/ngsi_v2/cb.py
```diff
@@ -1717,6 +1720,54 @@ def patch_entity(self,
...
+            the value is not none.
+            If the value is a list and any item is not none, the value is not none.
+            If it's neither dict nore list, bool is used.
+            """
+            if isinstance(value, dict):
+                return any([_value_is_not_none(value=_v) for _v in value.values()])
+            if isinstance(value, list):
+                return any([_value_is_not_none(value=_v)for _v in value])
+            else:
+                return bool(v)
```
djs0109, line 1750: return bool(value)?

filip/clients/ngsi_v2/cb.py
FWuellhorst: Mhm, depends, or? If I have an old subscription with key `e = None`, I want the new key `e =  5` to be an update. Using `or` would lead to `None == 5 -> true` ?!

djs0109: That's true. A `None` normally indicates that the parameter is unset and the default value is used. So I think it would be ok if this function gives a warning or even returns these unequal keys as results? How to deal with these keys will then be the job of specific use case.

FWuellhorst: Ok! Do you want to add a warning yourself and make the changes? Or should I do that?

djs0109: I will do it ;)

djs0109:
I think we still need a test for this new feature. @FWuellhorst Can you write a test for this new feature under `test_ngsi_v2_cb`? For example, the most basic test will be: try to post the same or slightly different `Subscription` twice.

djs0109:
@FWuellhorst sry for the delay, can you still remember the reason to define another test subscription, i.e. `sub_with_nans`? This current cause some tiny issue

FWuellhorst:
@djs0109 I think I had some subscription with nan values like `[]` which raised an issue. I can have a look at the logic and try to fix it, or do you want to?

djs0109:
@FWuellhorst I fixed the empty expression issue. But there is still some tiny issue remaining caused by the for loop. It would be great if you can have a look :)

tstorek:
@djs0109 @FWuellhorst There might be also tiny issue with the default values of the `Subscription` model. As @lidiawestphal notes in #181 in case that `attrs` is `none` or not included in the model and `explcitAttrs` is not present, Orion might assume an empty list for `attrs`. Hence, the server-subscription and the client subscription instantaneously are out of sync. That is also the reason why the `quick and dirty` variant with hasing the two json strings currently does not work. I am happy to help after my return. As registtrations need such a pattern as well :)

djs0109:
The issue of `notification.attrs=None` is now implemented as a testcase, which can currently pass. So I assume that this issue is already solved. However, before this branch can be merged. I still want some extra tests to be implemented in `tests/clients/test_ngsi_v2_cb.py`

djs0109:
Hi @RCX112, could you please complete this test here? I think after that we are ready to merge this PR
https://github.com/RWTH-EBC/FiLiP/blob/12789a0ede6dab99dfab2eadd9d91595048db9b3/tests/clients/test_ngsi_v2_cb.py#L637

RCX112:
Hey @djs0109, I completed the test function in https://github.com/RWTH-EBC/FiLiP/blob/6155f430f5ad67a0382948dbb27bdc368c3937ec/tests/clients/test_ngsi_v2_cb.py#L534

where I added two extra subscriptions and also tested for duplicates. During the test I made some minor changes in the check function between two subscription dicts https://github.com/RWTH-EBC/FiLiP/blob/6155f430f5ad67a0382948dbb27bdc368c3937ec/filip/clients/ngsi_v2/cb.py#L1793

As an example if I have `condition1 = None` and `condition2 = {'attrs': [], 'expression': None}`, they should be both equal. I also ignored the attribute `timesSent`, because if the rest of the subscription dicts are equal I don't want to start counting from a second subscription, which is identical: https://github.com/RWTH-EBC/FiLiP/blob/6155f430f5ad67a0382948dbb27bdc368c3937ec/filip/clients/ngsi_v2/cb.py#L1841C1-L1845C29

djs0109:
@RCX112 thank you, that looks good! I just slightly adjusted the code format. I will merge after the test run though

djs0109 (approved):

FWuellhorst:
Thanks for finalizing this :)

PR #142, opened 2022-04-22
fix: missing parameter handling for entity_type in update_attribute_value

cl0-de:


tstorek (approved):
Thank you for this fix :)

Issue #141, opened 2022-04-11
Get all subscriptions for an entity


sbanoeon:
**Is your feature request related to a problem? Please describe.**
The goal is to retrieve a list of subscriptions corresponding to entity id and entity type.

**Describe the solution you'd like**
The function "get_subscriptions_by_entity" would return subscriptions of which entity_id matches the passed id AND entity_type matches the passed type (also based on id and type pattern flags).

tstorek:
@sbanoeon I think that this is related to #137 I suggest to add a function like 
`def is_subset(self, other)` to the subscription subclass `EntityPattern`. This would also help for detecting if a subscription might be already covered by another subscription. Anyway from  the API specs. one would always need to get the whole subscription list first, I think. @FWuellhorst what do you think about combining these two by adding `__eq__` to he models where possible and the proposed function where patterns may be required , 

Anyway I am happy about pull requests. @sbanoeon please check on the other pull requests :)

Thanks!

tstorek:
@FWuellhorst i am fine with using the word matches or equals. The advantage of __eq__ is the automatic recursive calling. I am fine with both i think. Anyway i like it better than having a utility for this :)

PR #147, opened 2022-05-25
feat: added function get_subscriptions_by_entity with test

sbanoeon:
For #141

### DIFF
--- a/filip/utils/filter.py
+++ b/filip/utils/filter.py
@@ -1,0 +1,43 @@
+"""
+Filter functions to keep client code clean and easy to use.
+"""
+from typing import List
+from filip.clients.ngsi_v2 import ContextBrokerClient
+from filip.models import FiwareHeader
+from filip.models.ngsi_v2.subscriptions import Subscription
+
+
+def filter_subscriptions_by_entity(entity_id: str,
+                                   entity_type: str,
+                                   url: str = None,
+                                   fiware_header: FiwareHeader = None,
+                                   subscriptions: List[Subscription] = None,
+                                   ) -> List[Subscription]:
+    """
+    Function that filters subscriptions based on the entity id or id pattern
+    and entity type or type pattern. The function can be used in two ways,
+    wither pass list of subscriptions to filter based on entity or directly pass
+    client information to filter subscriptions in a single request.
+    Args:
+        entity_id: Id of the entity to be matched
+        entity_type: Type of the entity to be matched
+        url: Url of the context broker service
+        fiware_header: Fiware header of the tenant
+        subscriptions: List of subscriptions to filter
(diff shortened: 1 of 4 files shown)


tstorek:
@sbanoeon thank you for this contribution. I think this is related to #141 and #137. 
Please, check if this should not better become a filter function in utils. Since it actually copies the code from `get_subscriptions_list`. Hence, let's rather prepare the filter function that can be used for filtering matching subscriptions on multiple levels from entity matching down to complete logic matching :)

filip/utils/filter.py
```diff
@@ -0,0 +1,43 @@
...
+                                   url: str = None,
+                                   fiware_header: FiwareHeader = None,
+                                   subscriptions: List[Subscription] = None,
+                                   ) -> List[Subscription]:
+    """
+    Function that filters subscriptions based on the entity id or id pattern
+    and entity type or type pattern. The function can be used in two ways,
+    wither pass list of subscriptions to filter based on entity or directly pass
+    client information to filter subscriptions in a single request.
+    Args:
```
tstorek, line 21: missing blankline

tstorek (changes requested):
@sbanoeon looks quite good. please add the corresponding test

tests/utils/test_filter.py
```diff
@@ -0,0 +1,10 @@
+"""
```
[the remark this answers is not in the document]
sbanoeon: You mean test_subscriptions.py, right?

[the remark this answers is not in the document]
tstorek: oh sorry. I meant both the test file and the module in filip.utils.

tests/utils/test_filter.py
```diff
@@ -2,9 +2,86 @@
...
+from filip.clients.ngsi_v2 import ContextBrokerClient
+from filip.models.base import FiwareHeader
+from filip.models.ngsi_v2.base import EntityPattern
+from filip.models.ngsi_v2.subscriptions import Subscription
+from filip.utils import filter
+from filip.utils.cleanup import clear_all
+from tests.config import settings
 
 
 class TestFilterFunctions(unittest.TestCase):
```
tstorek, line 16: please have a look at utils.iot there you will find the other filter I mentioned

tstorek (changes requested):
@sbanoeon looks quite good. please. make the last changes :)

tests/utils/test_filter.py
sbanoeon: Then I think the best place for this testcase is in `tests.clients.test_ngsi_v2_cb.py`. What do you think? Or it should be separate in `tests.utils`?

Issue #148, opened 2022-06-17
Throttling=0 is not working when posting a subscription


StWiemann:
**Describe the bug**
When creating a subscription with ql_client.post_subscription() a value < 1 is not accepted. This leads to a throttling of min. 1.

**To Reproduce**
1. e.g. "TimeSeries Data with QL"-Example. Only two isntead of six values will end up in the DB.

**Expected behavior**
Throttling 0 should be possible to allow data-collection sub 1s.
 
This is a bug in the quantumleap-client. Function "post_subscription" [in Line 211](https://github.com/RWTH-EBC/FiLiP/blob/d24f47daa272a65ccf9c92522374bc5228b9a3d1/filip/clients/ngsi_v2/quantumleap.py#L211).

`if throttling` will evaluate to `false` when using "0".
Also, I think the following if-statement should be `if throttling <0`.

I can do a PR for this or you just fix this with the next push :)

Cheers

tstorek:
@StWiemann thank you for raising an issue about this. Hope you enjoy the library so far. Actually, I did never use this function myself since I usually use the advanced subscription mechanism in the context broker. Please, feel free to create a PR. I will assign this to you. 
Please keep in mind that the branch will be created automatically. 

Additionally, please keep in mind to double-check quantum leap docs on this and write a simple test case. 

Thank you very much!

PR #149, opened 2022-06-30
fix: Throttling=0 is not working when posting a subscription

tstorek:
@StWiemann I added this code snippet. Please have a look if this works fine for you!
It would be great if you could write a quick test for it!

For #148


StWiemann:
Hi @tstorek, 
sorry that I didn't get around to push the proposed PR last week. This looks almost identical to what I've done and works just fine! Thank you.

tstorek:
@StWiemann would you be so kind to write a test for this?

StWiemann:
@tstorek Yes, would you like me to add this to clients -> `class TestTimeSeries` in the `test_ngsi_v2_timeseries.py` ?

tstorek:
@StWiemann sounds good to me. Simply use my branch for it. Please

djs0109:
I don't know, if you also notice the returned message from QL: `This API is deprecated, it will be removed in version 0.9`. It seems like this kind of request will no longer be possible in the future.

tstorek:
@djs0109 good point. I will move the test to the contextbroker ;) same same :)

djs0109 (approved):

Issue #151, opened 2022-08-10
`delete_entity` should receive proper IoTAClient object


tstorek:
**Describe the bug**
In the contextbrokerclient the delete_entity function an iota_url to delete related devices.
This leads to automated creation of an IOTAClient. 
However, this client will simple inherit headers from the contextbroker client, which can lead to unwanted behavior.

This also affects the corresponding bulk operation!

**To Reproduce**
1. Create a Device connected to a ContextEntity using an IoTA-Client
2. Delete the ContextEntity via ContextBrokerClient 
3. Watch the absence of header of the intermediate client.

**Expected behavior**
To make functionality more transparent a complete client should be passed to the function.
We can leave the current args for now and put a deprecation warning.

**Environment (please complete the following information):**
 - Version 0.2.2

Issue #143, opened 2022-05-06
Unrelevant devices get deleted by cb.delete_entity()


djs0109:
**Describe the bug**
cb.delete_entity( ) has a flag "delete_devices". As described, if "delete_devices" is True, the devices that reference the specified entity should be deleted. However, in reality, all devices with the same entity_type are deleted.

**To Reproduce**
1. Create some devices with the same entity_type, e.g. TemperatureSensor
```
[
        {
            "device_id": "device:001",
            "apikey": "n5gehtutorial",
            "service": "workshop",
            "service_path": "/",
            "entity_name": "urn:ngsi-ld:TemperatureSensor:001",
            "entity_type": "TemperatureSensor",
            "transport": "MQTT",
            "attributes": [
                {
                    "object_id": "temp",
                    "name": "temperature",
                    "type": "Number"
                }
            ],
            "lazy": [],
            "commands": [],
            "static_attributes": [],
            "internal_attributes": [],
            "protocol": "IoTA-JSON",
            "explicitAttrs": false,
            "ngsiVersion": "v2"
        },
        {
            "device_id": "device:002",
            "apikey": "n5gehtutorial",
            "service": "workshop",
            "service_path": "/",
            "entity_name": "urn:ngsi-ld:TemperatureSensor:002",
            "entity_type": "TemperatureSensor",
            "transport": "MQTT",
            "attributes": [
                {
                    "object_id": "temp",
                    "name": "temperature",
                    "type": "Number"
                }
            ],
            "lazy": [],
            "commands": [],
            "static_attributes": [],
            "internal_attributes": [],
            "protocol": "IoTA-JSON",
            "explicitAttrs": false,
            "ngsiVersion": "v2"
        },
        {
            "device_id": "device:003",
            "apikey": "n5gehtutorial",
            "service": "workshop",
            "service_path": "/",
            "entity_name": "urn:ngsi-ld:TemperatureSensor:003",
            "entity_type": "TemperatureSensor",
            "transport": "MQTT",
            "attributes": [
                {
                    "object_id": "temp",
                    "name": "temperature",
                    "type": "Number"
                }
            ],
            "lazy": [],
            "commands": [],
            "static_attributes": [],
            "internal_attributes": [],
            "protocol": "IoTA-JSON",
            "explicitAttrs": false,
            "ngsiVersion": "v2"
        }
]
```
2. Try to delete the entity "urn:ngsi-ld:TemperatureSensor:001" with cb.delete_entity()
[...]

PR #152, opened 2022-08-10
fix: delete entity should receive proper io ta client object

tstorek:
#141 #143 #151


tstorek:
@djs0109 please double check if everything works as expected :)

djs0109:
LGTM
Btw, I can not pass two tests in `tests/clients/test_ngsi_v2_cb.py`: `test_mqtt_subscriptions` and `test_patch_entity`, but I think these are not relevant for this issue

tstorek:
@djs0109 `tests.clients.test_ngsi_v2_cb.test_mqtt_subscriptions` should not be affected as it does not make use of the IoTAgent in any way. same for `test_patch_entity`

djs0109 (approved):

sbanoeon (approved):

PR #163, opened 2022-09-20
fix: remove additional escape chars when sending a string

windrad6:
Fix issue with string format. The old version is escaped with \ . This results in an error when sending a string value to fiware


tstorek:
@windrad6 Thank you. For the request. If possible a simple test in that you send the string and fetch again, would be awesome. :)

windrad6:
Test case is added

tstorek (approved):
excellent thank you!

PR #171, opened 2022-11-08
169 service group can not be retrieved by apikey and resource

### LINKED ISSUE #169: Service Group can not be retrieved by apikey and resource
IoT Agent endpoint does not support any parameters for querying service group, see [here](https://iotagent-node-lib.readthedocs.io/en/latest/api.html#get-iotservices). Therefore, the FiliP function [get_group](https://github.com/RWTH-EBC/FiLiP/blob/526f55d2261dda0fc1dfc2780eef790b75df0140/filip/clients/ngsi_v2/iota.py#L156) does not work as expected. We can implement some workaround, e.g. filter function...

djs0109:
close #169

### DIFF
--- a/filip/clients/ngsi_v2/iota.py
+++ b/filip/clients/ngsi_v2/iota.py
@@ -93,7 +93,8 @@ class IoTAClient(BaseHttpClient):
         url = urljoin(self.base_url, 'iot/services')
         headers = self.headers
         data = {'services': [group.dict(exclude={'service', 'subservice'},
-                                        exclude_defaults=True) for
+                                        exclude_none=True,
+                                        exclude_unset=True) for
                              group in service_groups]}
         try:
             res = self.post(url=url, headers=headers, json=data)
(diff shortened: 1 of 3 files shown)


filip/clients/ngsi_v2/iota.py
```diff
@@ -163,19 +164,15 @@ def get_group(self, *, resource: str, apikey: str) -> ServiceGroup:
...
-        except requests.RequestException as err:
-            self.log_error(err=err, msg=None)
-            raise
+        groups = self.get_group_list()
+        groups = filter_group_list(group_list=groups, resources=resource, apikeys=apikey)
+        if len(groups) == 1:
+            group = groups[0]
+            return group
+        elif len(groups) == 0:
+            raise KeyError(f"Service group with resource={resource} and apikey={apikey} was not found")
```
djs0109, line 173: I am not sure with the KeyError here. Maybe there is a better Error type or even another way to inform the users.

tstorek: Well, you can use a ValueError, but in the very end you are looking for a specific combination of keys ;)

djs0109: Thats true. Would it be better, if we build a custom error type here? Something like "GroupNotFoundError".

tstorek: @djs0109 I suggest to stick to the keyError

djs0109: OK, then I will resolve this point ;)

sbanoeon (approved):

PR #187, opened 2023-03-03
chore: add extra config to make Unit model hashable

### LINKED ISSUE #186: Make unit model hashable
Currently, the units model is not hashable although the content should be immutable and, hence, frozen. 
But due to the missing Config in the model the `__hash__` function is not generated by pydantic.

I would like to add the following content to the model:

```python
class Unit(BaseModel):
    ...

    class Config:
        extra = 'ignore'
        allow_population_by_field_name = True
        allow_mutation = False
        frozen = True
           
```

**Describe alternatives you've considered**
Inherit, the Unit model and add a specific Config object. However, this may lead to inconssitency

**Additional context**
Use case: 
I have a list ob sensors that also contains information about the unit. 
For mapping measurements to the sepcific sensors I regularly must loop throuth this list. Therefore, caching such lookup can be wuite useful. Caching though requires hashing of the Objects in the list that I qant to operate on.

djs0109:
close #186

### DIFF
--- a/filip/models/ngsi_v2/units.py
+++ b/filip/models/ngsi_v2/units.py
@@ -136,6 +136,8 @@ class Unit(BaseModel):
     class Config:
         extra = 'ignore'
         allow_population_by_field_name = True
+        allow_mutation = False
+        frozen = True
 
     @root_validator(pre=False, allow_reuse=True)
     def check_consistency(cls, values):


djs0109:
@tstorek I have made the changed you suggested in this branch, can you add a test for caching under `test_units.py`?

tstorek:
@djs0109 I added the test you requested. Including the example how lru_cache works ;) it raises perfomrance a lot  with repeating function calls

djs0109:
Thanks for providing this example 👍 I can see a speed boost of about 100-500 times

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+#### v0.2.5
+- fixed inconsistency of `entity_type` as required argument ([#188](https://github.com/RWTH-EBC/FiLiP/issues/188))
+
```
tstorek, line 3: please update before merge or do you autogenerate from here on? Probably better

filip/semantics/semantics_models.py
```diff
@@ -1132,7 +1132,7 @@ def remove(self, v):
         elif isinstance(v, SemanticIndividual):
             self._set.remove(v.get_name())
         else:
-            raise KeyError
+            raise KeyError(f"v is neither of type SemanticIndividual nor SemanticClass but {type(v)}")
```
tstorek, line 1135: The error use different code patterns

tstorek (changes requested):
only minor things

tests/models/test_units.py
```diff
@@ -64,7 +64,7 @@ def test_unit_model_caching(self):
 
         self.assertEqual(unit.__hash__(), unit.__hash__())
 
-        @functools.lru_cache
+        @functools.lru_cache(maxsize=128)
```
tstorek, line 67: isn't that the default anyway? I don't think that python 3.7 should fail because of this. https://docs.python.org/3.7/library/functools.html

djs0109: I also wonder about that, but it actually fails with the error message:
```
TypeError: Expected maxsize to be an integer or None
```
btw, the python examples all give the `maxsize` explicitly.

tstorek:
@djs0109 is this one fiex now? Probabably, we should double check by merging the main branch into it :)

djs0109:
@tstorek it should have been fixed. But you are right, we should double check, because it is implemented before the pydantic migration. I will try to solve the conflict next week

PR #188, opened 2023-03-03
chore: make `entity_type` optional

### LINKED ISSUE #168: Inconsistent use if entity_type parameter
**Describe the bug**
The parameter entity_type is used inconsistently in different functions in cb.py. For get_entity_attributes it is an optional parameter, whereas for delete_entity it is a required parameter. It should either be one or the other. Since one could argue that fiware requires the combination of entity_id and entity_type to be unique, both should be required in all functions that access an entity. On the other hand, always providing an entity_type leads to unnecessary code complexity and could in a lot of cases be dismissed. 

http://fiware.github.io/specifications/ngsiv2/stable/

"Context EntitiesContext entities, or simply entities, are the center of gravity in the FIWARE NGSI information model. An entity represents a thing, i.e., any physical or logical object (e.g., a sensor, a person, a room, an issue in a ticketing system, etc.). Each entity has an entity id.Furthermore, the type system of FIWARE NGSI enables entities to have an entity type. Entity types are semantic types; they are intended to describe the type of thing represented by the entity. For example, a context entity with id sensor-365 could have the type temperatureSensor.Each entity is uniquely identified by the combination of its id and type."

djs0109:
close #168
@windrad6 As we previously discussed, we will make the `entity_type` not required for most of the functions


djs0109:
@tstorek you are right, there were still some works to do. It should be ready for review now

tstorek:
> @tstorek you are right, there were still some works to do. It should be ready for review now

@djs0109 Thank you. Unfortunately, I will not be available before march. Maybe you'll find some substitution

djs0109:
@tstorek no worry, I will then ask other reviewers ;)

filip/clients/ngsi_v2/cb.py
```diff
@@ -598,25 +601,29 @@ def delete_entity(self,
...
-                if device.entity_type == entity_type:
-                    iota_client_local.delete_device(device_id=device.device_id)
+                for device in iota_client_local.get_device_list(
+                        entity_names=[entity_id]):
+                    if device.entity_type == entity_type:
+                        iota_client_local.delete_device(device_id=device.device_id)
 
-            iota_client_local.close()
+                iota_client_local.close()
+            else:
```
sbanoeon, line 624: Why should it not be possible to delete a device without providing entity_type?

djs0109: I think the idea is that only the combination of `entity_name` and `entity_type` can uniquely identify an entity. But this is a decision we can choose

djs0109: It will be revised to delete all devices matched with that "id"

sbanoeon (approved):

PR #193, opened 2023-05-30
get() method of Units dose not work properly

### LINKED ISSUE #189: false string matching `__getitem__` of Units
**Describe the bug**
Currently, the  `__getitem__` method of Units class does not work correctly. This due to worng string matching in the
the line where, the function tries to find the corrrect name in the database.

**To Reproduce**

```python
from filip.models.ngsi_v2.units import Units

units = Units()
print(units.get("degree Celsius"))
```

raises an error with unit suggestions comming from `Unit.__getitem__`

**Expected behavior**
This should simply work. As using the Unit-Model itself. But that one uses a different lookup function and not the combined one of the `Units().get(item)`

**Suggested sollution**

Change line 235 to 

```python

idx = self.units.index[((self.units.CommonCode == item.upper()) |
                        (self.units.Name.str.casefold() == item.casefold()))]
```

or even better combine the lookups functions

djs0109:
close #189

### DIFF
--- a/filip/models/ngsi_v2/units.py
+++ b/filip/models/ngsi_v2/units.py
@@ -233,7 +233,7 @@ class Units:
             Unit
         """
         idx = self.units.index[((self.units.CommonCode == item.upper()) |
-                                (self.units.Name == item.casefold()))]
+                                (self.units.Name.str.casefold() == item.casefold()))]
         if idx.empty:
             names = self.units.Name.tolist()
             suggestions = [item[0] for item in process.extract(


tstorek (changes requested):
Seems correct. Maybe we should add a test for it as well using the line from the orginal issue ?

tstorek:
@djs0109 I added the test if you are ine with it. we can merge. Please trigger the pipeline again

tstorek (approved):

tstorek (approved):

tstorek (approved):

PR #200, opened 2023-07-19
Added version checker for correct cb version #190

### LINKED ISSUE #190: Warn about wrong orion version
When using the latest filip with older fiware versions, some errors occur. For instance, `overwrite_metadata` in `update_attribute` is not supported and returns dubios http errors.

Upon opening the `ContextBrokerClient`, check the version of the context broker and compare it to the supported versions of filip.
Maybe add a list with links between fiware and filips versions that are compatible.

**Describe alternatives you've considered**
Warn about not using the latest fiware versions upon starting instances of `ContextBrokerClient` or IOTA etc.

djs0109:
close #190


filip/clients/ngsi_v2/cb.py
```diff
@@ -157,6 +160,22 @@ def get_version(self) -> Dict:
             self.logger.error(err)
             raise
 
+    def _check_correct_cb_version(self) -> None:
+        """
+        Checks whether Orion version is >= 3.6.0, since breaking change was introduced there which introduced new
+        option, which is implemented in FiLiP, but won't work with lower orion versions.
+        """
+        orion_version = self.get_version()['orion']['version']
+        if version.parse(orion_version) < version.parse('3.6.0'):
```
djs0109, line 169: I think some significant changes have also been made in `3.8.0`. @tstorek Do you think we should also set an upper limit for the supported Orion version? like version < 3.8.0

tstorek: @djs0109 @jkriwet I think that it is worth having a look at the changes: https://github.com/telefonicaid/fiware-orion/releases. If introducing such bounds would also mean creating the issues to support the features in coming versions.

sbanoeon:
@djs0109 I have made the necessary changes. Could you please review it?

filip/config.py
```diff
@@ -65,9 +65,7 @@ class Settings(BaseSettings):
         default="filip", validation_alias=AliasChoices("FIWARE_SERVICE")
     )
 
-    MINIMUM_ORION_VERSION: str = Field(
-        default="3.6.0", validation_alias=AliasChoices("MINIMUM_ORION_VERSION")
-    )
+    MINIMUM_ORION_VERSION: str = "3.6.0"
```
sbanoeon, line 68: @djs0109 shouldn't it be a field? We can also pass it from the environment.

PR #208, opened 2023-11-02
Add authorization field to fiware-header

Malte96:
Simple way to use an authorization field via Baerer-Token in the fiware-header

### DIFF
--- a/filip/models/base.py
+++ b/filip/models/base.py
@@ -88,6 +88,20 @@ class FiwareHeader(BaseModel):
     class Config(BaseConfig):
         allow_population_by_field_name = True
         validate_assignment = True
+        
+class FiwareHeaderSecure(FiwareHeader):
+    """
+    Defines entity service paths and a autorization via Baerer-Token which are supported by the NGSI
+    Context Brokers to support hierarchical scopes:
+    https://fiware-orion.readthedocs.io/en/master/user/service_path/index.html
+    """
+    authorization: str = Field(
+        alias="authorization",
+        default="",
+        max_length=3000,
+        description="authorization key",
+        regex=r".*"
+    )
 
 
 class FiwareRegex(str, Enum):


tstorek:
@Malte96 Thanks, the contribution. Please ellaborate, why you would prefer this option over the standard `requests`-functionality.
The Fiware-Header actually only exists because Fiware-service and Fiware-servivepath are specific to FIWARE and `forced` to be used for interaction with the context-broker. `Authorization` via Bearer is a standard functionality. Keep in mind that FiLiP clients open a `request session`. You can add any headers you like to the clients. They all have a header-attribute.

Have a look here: https://ebc.pages.rwth-aachen.de/EBC_all/github_ci/FiLiP/master/docs/api/filip.clients.html#module-filip.clients.base_http_client

Simply try the following
```
from filip.clients.ngsi_v2 import ContextBrokerClient

# Host address of Context Broker
CB_URL = "http://localhost:1026"

fiware_header = FiwareHeader(service='filip', service_path='/example')

with ContextBrokerClient(url=http://localhost:1026, fiware_header=fiware_header) as cb_client:
        print(f"OCB Version: {cb_client.headers}")

```

Malte96:
Hello, of course the necessity of the FIWARE header is clear, and also the possibility of adding additional headers. Wouldn't it be useful, to include a mostly necessary authentication directly, so that adding the Baerer token is easier? 
If this is not part of the intention, would it make sense to add instructions for adding this header in the documentation?

tstorek:
@djs0109 What do you think. In the past I thought about showing some implementations. However, the difficulty was always that the authentification/authorization workflows (OAuthu2/Basic Auth/OIDC,2FA/MFA) may differ for different platform setups. Therefore, we cannot fully test for general use case. This would up to the developer of the individual use case. Maybe we can simple Link some general tutorials

djs0109:
I would also say that we can only provide a tutorials showing how to enable the authentication/authorization function with a very specific use case, e.g. OAuth. But since there are too many variants of the methods, we are not able to provide a comprehensive support.

tstorek:
@Malte96 Maybe you can add an example using parts of your implementation. That would be awesome!

Malte96:
@tstorek and @djs0109: I have added an example in the examples, can this be used?

djs0109:
@Malte96 Thank you! I just made a few small changes and solve the conflict with master branch. Could you also have a look, that it is still fulfill your requirements? Then I will approve the PR

Malte96:
Yes, looks good

djs0109 (approved):

PR #229, opened 2023-12-29
fix: remove exclude_defaults

### LINKED ISSUE #228: Requests remove defaults
**Describe the bug**
Currently, many requests, e.g., in ContextBroker-Client automatically ignore model-fields that have their default values.
This is independent from whether the value is explicitly set or not. Please, check on Pydantic docs:
https://docs.pydantic.dev/2.5/api/base_model/#pydantic.main.BaseModel.model_dump_json

This this hinders if one wants to inherit from, e.g,  ContextEntity and to define a strict model for a specific use case.

**To Reproduce**
```python
from pydantic import Field, BaseModel

class ModelA(BaseModel):
    """
    Model with a required field that has no default value
    """
    name: str = Field()

class ModelB(BaseModel):
    """
    Derived model with a field that has a default value
    """
    name: str = Field(
        default="something",
    )


if __name__ == "__main__":
    model = ModelA(name="something")
    print(model.model_dump_json(indent=2, exclude_defaults=True))
    """
    {}
    """
    model = ModelB()
    print(model.model_dump_json(indent=2, exclude_defaults=True))
    """
    {}
    """
    # Explicitly set a value that is equal to the default value
    model = ModelB(name="something")
    print(model.model_dump_json(indent=2, exclude_defaults=True))
    """
    {}
    """
    # Set a value that is not equal to the default value
    model = ModelB(name="another thing")
    print(model.model_dump_json(indent=2, exclude_defaults=True))
    """
    {
      "name": "another thing"
    }
    """
```

**Expected behavior**
This is not very user friendly as the user would expect the default value to be send to FIWARE if present in the model.

tstorek:
closes #228


tstorek:
@djs unfortunetaly, I cannot see, where the pipeline fails. Please, let me know, what the problem is :)

filip/clients/ngsi_v2/cb.py
```diff
@@ -1165,49 +1179,50 @@ def post_subscription(self,
...
+        url = urljoin(self.base_url, "v2/subscriptions")
         headers = self.headers.copy()
-        headers.update({'Content-Type': 'application/json'})
+        headers.update({"Content-Type": "application/json"})
         try:
             res = self.post(
                 url=url,
                 headers=headers,
-                data=subscription.model_dump_json(exclude={'id'},
-                                                  exclude_unset=True,
```
djs0109, line 1204: Why the `exclude_unset` is also removed here?

filip/clients/ngsi_v2/cb.py
```diff
@@ -219,59 +221,53 @@ def post_entity(self,
...
 
         """
-        url = urljoin(self.base_url, 'v2/entities')
+        url = urljoin(self.base_url, "v2/entities")
         headers = self.headers.copy()
         try:
             res = self.post(
-                url=url,
-                headers=headers,
-                json=entity.model_dump(exclude_unset=True,
```
djs0109, line 236: Why the `exclude_unset` is also removed here?

filip/clients/ngsi_v2/cb.py
```diff
@@ -1351,23 +1365,21 @@ def post_registration(self, registration: Registration):
...
+        url = urljoin(self.base_url, "v2/registrations")
         headers = self.headers.copy()
-        headers.update({'Content-Type': 'application/json'})
+        headers.update({"Content-Type": "application/json"})
         try:
             res = self.post(
                 url=url,
                 headers=headers,
-                data=registration.model_dump_json(exclude={'id'},
-                                                  exclude_unset=True,
```
djs0109, line 1362: Why the `exclude_unset` is also removed here?

filip/clients/ngsi_v2/cb.py
```diff
@@ -1530,23 +1546,24 @@ def query(self,
...
-            params['options'] = ','.join([response_format, 'count'])
+                raise ValueError(f"Value must be in {list(AttrsFormat)}")
+            params["options"] = ",".join([response_format, "count"])
         try:
-            items = self.__pagination(method=PaginationMethod.POST,
-                                      url=url,
-                                      headers=headers,
-                                      params=params,
-                                      data=query.model_dump_json(exclude_unset=True,
-                                                                 exclude_none=True),
```
djs0109, line 1548: Same as above

filip/clients/ngsi_v2/iota.py
```diff
@@ -96,9 +96,8 @@ def post_groups(self,
         url = urljoin(self.base_url, 'iot/services')
         headers = self.headers
         data = {'services': [group.model_dump(exclude={'service', 'subservice'},
-                                              exclude_none=True,
-                                              exclude_unset=True) for
```
djs0109, line 100: `exclude_unset`

filip/clients/ngsi_v2/quantumleap.py
```diff
@@ -115,9 +115,7 @@ def post_notification(self, notification: Message):
         headers = self.headers.copy()
         data = []
         for entity in notification.data:
-            data.append(entity.model_dump(exclude_unset=True,
```
djs0109, line 118: `exclude_unset`

djs0109:
@tstorek I just marked all the places where the `exclude_unset` is also removed. Did you have some criteria for removing or not removing the `exclude_unset`?

filip/clients/ngsi_v2/cb.py
tstorek: @djs0109 its because if it was excluded. The default values if not explitly set would be removed as well. Hence, the user would expect the default value will be sent to the plattform.

tstorek:
@djs0109 I explained above why I think that it is better to also exclude unset. This only means taht all basic filip models must use None as default, if they should not be sent by default. Sound reasonable?

djs0109:
@tstorek It sounds reasonable. But for the consistency, do you think it is better to remove all `exclude_unset`? Btw, the tests can run through now.

tstorek:
@djs0109 i removed them wherever it was safe at the first glance. Probably for consistency it would be good to remove them all. Please, feel free to check. :) thank you

djs0109:
@tstorek I remove all `exclude_unset` for consistency. And I think this is actually better, in the case of NGSI endpoints changing certain behaviors, such as change the previous default value

djs0109 (approved):

PR #233, opened 2024-01-17
232  post entity will not completely override the existing entity

### LINKED ISSUE #232: `post_entity` will not completely override the existing entity
**Describe the bug**
`post_entity` will not completely override the existing entity

**To Reproduce**
1. create entity
```python
entity = {
    "id": "Test:001",
    "type": "Test",
    "humidity": {
        "type": "Number",
        "value": 1,
        "metadata": {}
    },
    "temperature": {
        "type": "Number",
        "value": 1,
        "metadata": {}
    }
}
cb_client.post_entity(ContextEntity(**entity))
```
2. delete an attribute and update entity with `post_entity`
```python
entity_delete_attr = {
    "id": "Test:001",
    "type": "Test",
    "temperature": {
        "type": "Number",
        "value": 1,
        "metadata": {}
    }
}
cb_client.post_entity(ContextEntity(**entity_delete_attr ), update=True)
```
**Expected behavior**

The attribute `humidity` is deleted, which is currently not.

**Environment**
- Orion 3.7.0

djs0109:
closes #232

### DIFF
--- a/tests/clients/test_ngsi_v2_cb.py
+++ b/tests/clients/test_ngsi_v2_cb.py
@@ -11,6 +11,8 @@ import uuid
 
 import paho.mqtt.client as mqtt
 from datetime import datetime, timedelta
+
+import requests
 from requests import RequestException
 from filip.models.base import FiwareHeader
 from filip.utils.simple_ql import QueryString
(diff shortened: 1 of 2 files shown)


djs0109:
@tstorek by checking this issue, I found that the `patch_entity()` function is very heavy. And its behavior is more like a `PUT` request. Do you have an idea about that?

tstorek:
@djs0109 The reason for `patch_entity` is that Orion does not support partial updates. (Maybe using complex query). Hence, we introduced this one to allow for the functionality. Hence, it should merge both contents. And only overwriting those that should be updated or appended.

djs0109:
@tstorek By "partially update," do you mean like to send only a part of the attributes in the payload, and then Orion will only update the included attributes? If not, it would be great if you can add a simple implementation here
https://github.com/RWTH-EBC/FiLiP/blob/d6068183ea4794e98900b996e44d2cbbfcf5b9da/tests/clients/test_ngsi_v2_cb.py#L244

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -229,6 +231,147 @@ def test_entity_operations(self):
...
+                entity_post.update_attribute(attrs=[attr_append])
+                client.post_entity(entity=entity_post, patch=True)
+                self.assertEqual(client.get_entity(entity_id=entity_post.id),
+                                 entity_post)
+                # 3) delete attribute
+                entity_post.delete_attributes(attrs=[attr_append])
+                client.post_entity(entity=entity_post, update=True)
+                self.assertEqual(client.get_entity(entity_id=entity_post.id),
+                                 entity_post)
+
```
RCX112, line 294: @djs0109 After each step (so after `post`, `update_entity `and `override_entity`) I would clear the OCB and reset the variable `attr_append`:

```
clear_all(fiware_header=self.fiware_header, cb_url=settings.CB_URL)
attr_append = NamedContextAttribute(**{
    "name": 'pressure',
    "type": 'Number',
    "value": 1050})
```

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -229,6 +231,147 @@ def test_entity_operations(self):
...
+            # update_entity()
+            if "update_entity":
+                client.post_entity(entity=entity_init, update=True)
+                entity_update = entity_init.model_copy(deep=True)
+                # 1) append attribute
+                entity_update.add_attributes(attrs=[attr_append])
+                with self.assertRaises(requests.RequestException):
+                    client.update_entity(entity=entity_update, append_strict=True)
```
RCX112, line 302: @djs0109 This call still makes an update to the entity. So we could add the follwing assertion:

```
self.assertEqual(client.get_entity(entity_id=entity_update.id),
                                 entity_update)
```

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -229,6 +231,147 @@ def test_entity_operations(self):
...
+                    "type": 'Number',
+                    "value": 2050})
+                entity_patch.update_attribute(attrs=[attr_append])
+                client.patch_entity(entity=entity_patch)
+                self.assertEqual(client.get_entity(entity_id=entity_patch.id),
+                                 entity_patch)
+                # 3) delete attribute
+                entity_patch.delete_attributes(attrs=[attr_append])
+                client.patch_entity(entity=entity_patch)
+                # TODO It is expected here not to pass the test
```
RCX112, line 372: @djs0109 The patch function can also delete attributes from an entity with the DELETE request

RCX112 (commented):
There are only some small comments from my side :)

tests/clients/test_ngsi_v2_cb.py
djs0109: Good call. I add the `clear_all`, and use `attr_append_update` for updating attribute, so that the `attr_append` will not be touched during the whole test.

tests/clients/test_ngsi_v2_cb.py
djs0109: You are right, I thought that the request will be completely blocked

tests/clients/test_ngsi_v2_cb.py
djs0109: Yes. Currently it is possible. But right now `patch_entity` works more like a `put_entity`. But I will remove the todos here and maybe open another issue

djs0109:
@tstorek It seems that the `patch` end points are supported now in orion v2 check the section `PATCH /v2/entities/{entityId}/attrs` in https://fiware-orion.readthedocs.io/en/master/orion-api.html . If I understand correctly, the real functionality of `patch` is implemented in `update_existing_entity_attributes`. Do you think it is a good idea to rename the methods to avoid missunderstanding? But the decision should be irrelevant for this issue

RCX112:
@djs0109 I have only one small comment on the line https://github.com/RWTH-EBC/FiLiP/blob/b6c1d76199419747ae46296bb4a0ca0049bf12b3/tests/clients/test_ngsi_v2_cb.py#L221
because it does not do very much. You can maybe make an assertion out of it like
`self.assertEqual(res_entity, client.get_entity(entity_id=self.entity.id, attrs=['temperature']))`. I think we can merge after this.

djs0109:
@RCX112 Thank you for the suggestion! But I think that will make the test a bit more hard-coded, because if we later add another initial attribute (e.g. "pressure") in `self.entity`, the code you suggested will not work. What do you think? If you think this point is critical, I can still change it.

RCX112:
@djs0109 Maybe you could change the line to something like this pseudo code:
`self.assertEqual(res_entity, client.get_entity(entity_id=self.entity.id, attrs=get_all_attributes_of_self.entity))`. This should be more flexible. If we keep the line as it is, this code only test the excecution without looking at the result.

djs0109:
@RCX112 Fair point :) I added it like that

RCX112 (approved):
@djs0109 I think the branch is ready to merge :)

Issue #220, opened 2023-12-05
ngsi-ld testcases for ld data models


mwr-ebc:
* Create testcases for all data models

PR #234, opened 2024-01-23
221 ngsi ld migrate v2 datamodels in ld

### LINKED ISSUE #221: Migrate NGSI-V2 datamodels for NGSI-LD integration
In order to integrate the NGSI-LD datamodel creation, implement additional logic to update the V2 datamodels regarding the migration steps defined by fiware:
https://fiware-datamodels.readthedocs.io/en/stable/ngsi-ld_howto/

djs0109:
close #221 #220

### DIFF
--- a/filip/models/ngsi_ld/context.py
+++ b/filip/models/ngsi_ld/context.py
@@ -732,10 +732,15 @@ class ContextLDEntity(ContextLDEntityKeyValues):
         Returns: context of the entity as list
 
         """
+        found_list = False
         for key, value in self.model_dump(exclude_unset=True).items():
             if key not in ContextLDEntity.model_fields:
                 if isinstance(value, list):
+                    found_list = True
                     return value
+        if not found_list:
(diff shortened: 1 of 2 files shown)


LaraRoth:
Will the attribute functions for NGSI-LD be implemented? Or will we not use them at all?

Maghnie (commented):
Very cool addition!

Some general questions for my understanding:

- Why is it necessary to have distinct named and unnamed versions? E.g. `NamedContextProperty` and `ContextProperty`. I would assume that a named version is preferable and that a name would be created by default, if not given by the user. 
- This is not limited to the LD branch in filip, but is the prefix "Context" necessary? E.g. couldn't we just say `NamedProperty` or `Property`, since a context is anyway implied when we use fiware?

djs0109:
The data models have been mostly implemented. The fine-tuning will be done together with the endpoints. To this end, this PR is closed.

PR #236, opened 2024-01-30
feat: add forcedUpdate and missing overrideMetadata

### LINKED ISSUE #235: Full support of options parameter for POST/PATCH/PUT entities
As listed in https://fiware-orion.readthedocs.io/en/master/orion-api.html#update-attribute-value-put-v2entitiesentityidattrsattrnamevalue. There are the following options parameters. For example, `flowControl` and `forcedUpdate` are currently not implemented in Filip.

| Options          | Description                                                                                                             |
|------------------|-------------------------------------------------------------------------------------------------------------------------|
| keyValues        | When used, the response payload uses the keyValues simplified entity representation. See Simplified Entity Representation section for details. |
| append           | Force an append operation.                                                                                              |
| overrideMetadata | Replace the existing metadata with the one provided in the request. See Metadata update semantics section for details.  |
| forcedUpdate     | Update operation have to trigger any matching subscription, no matter if there is an actual attribute update or no instead of the default behavior, which is to updated only if attribute is effectively updated. Check also the entityUpdate alteration type for the same effect. |
| flowControl      | Enable flow control mechanism, to avoid saturation under high-load scenarios. It is explained in this section in the documentation. |

djs0109:
closes #235


djs0109:
- [ ] Add test to test the functionality of `forcedUpdate` with notification

djs0109:
@j-pick I have implemented the test for the new feature. You could review it now :)

j-pick (approved):
Looks good to me :)

PR #240, opened 2024-02-06
chore: remove 37 and add 311 to test

### LINKED ISSUE #211: Python version upgrade to 3.10, 3.11, (3.12)
Currently, the community tends to use python `3.10`, `3.11`, and even `3.12` (see #209 #207 ). But `filip` can now officially only support python `3.7` - `3.9`. Therefore, an upgrade should be carefully taken to extend the python compatibility of `filip`

ToDos:
- [ ] Check `requirement.txt ` carefully and make certain adjustment, e.g. `igraph `
- [ ] Setup the test environment for Unitest in CI/CD, make sure all test cases can pass
- [ ] Add information in `README`, maybe on the header

djs0109:
closes #211


djs0109:
We currently cannot support `312` in CI/CD tests

djs0109:
@RCX112 it seems like, there are not changes required for the upgrading

RCX112:
@djs0109 Yes, I checked all the packages for Python 3.11 and 3.12. The only critical ones are `igraph~=0.9.8` and `matplotlib~=3.5.3`. In the Image you use in the CI-Pipeline `igraph~=0.9.8` will not be installed and I guess `matplotlib~=3.5.3` works in this case, because an Image like `python:3.11` is used (but for example it doesn't work for the Image `python:3.11-slim` or `python:3.11-alpine`). In general we should be careful with these two versions when using Python 3.11 and 3.12. The best and easiest solution is to update both packages to a newer version when using 3.11 and 3.12. For example `igraph~=0.11.3` and `matplotlib~=3.8.2` works pretty fine for me.

RCX112:
The tests also run through for 3.11 and 3.12 but I think they don't depend on these two packages

RCX112 (commented):
https://github.com/RWTH-EBC/FiLiP/pull/240#issuecomment-1943344582

djs0109:
Actually I have added `semantics` as an option for installation. Therefore `igraph` will only be installed when users explicitly specify, see [here](https://github.com/RWTH-EBC/FiLiP?tab=readme-ov-file#install-semantics-module-optional). And so do `matplotlib`. @RCX112 How do you install `filip`?  In our test environment, we use `pip install -r requirements.txt`. Therefore, all dependencies in `requirement.txt` are installed.

RCX112:
@djs0109 For testing I installed all the packages with `pip install -r requirements.txt`.

PR #256, opened 2024-03-05
225 make context entity more customizable

tstorek:


tstorek:
@djs0109 I added a more customizable implementation of the ContextAttribute and ContextEntity class. If the test all run trough I will add the examples and additional tests for inheritance options.

tstorek:
@djs0109 please let me know where the pipeline fails. Thus, I can double check. I only checked the offline-model-tests

djs0109:
@tstorek four tests have failed. I dont know if this information is enough for debugging 
```
FAILED tests/clients/test_ngsi_v2_cb.py::TestContextBroker::test_attribute_operations - TypeError: <class 'list'> does not match DataType.STRUCTUREDVALUE
FAILED tests/clients/test_ngsi_v2_iota.py::TestAgent::test_patch_device - TypeError: <class 'str'> does not match DataType.STRUCTUREDVALUE
FAILED tests/clients/test_ngsi_v2_iota.py::TestAgent::test_update_device - TypeError: <class 'str'> does not match DataType.STRUCTUREDVALUE
FAILED tests/models/test_base.py::TestModels::test_unit_as_metadata - TypeError: <class 'float'> does not match DataType.STRUCTUREDVALUE
```
Can you give me you the email address of you github account? Then I can grand you the right to the pipeline

djs0109:
> @djs0109 please let me know where the pipeline fails. Thus, I can double check. I only checked the offline-model-tests

Sounds good!

djs0109:
> @djs0109 I added a more customizable implementation of the ContextAttribute and ContextEntity class. If the test all run trough I will add the examples and additional tests for inheritance options.

Nice, that sounds good 👍

tstorek:
@djs0109 Okay, that sounds like minor issues. I'll send you may github-mail in a pm

tstorek:
@djs0109 I think that the pipeline works now. Do you want me to add additonal tests and examples? Where should I put the example?

djs0109:
@tstorek examples would be really helpful. How about under [examples/ngsi_v2](https://github.com/RWTH-EBC/FiLiP/tree/master/examples/ngsi_v2)?

tstorek:
@djs0109 apparently, there is one more failing check. Could you ellaborate agin, where? Maybe it would be quicker to fix it your self by putting the correct type. I also added an example know.

Not sure about tests tough.

djs0109:
@tstorek here is the error:
```
self = <tests.clients.test_ngsi_v2_iota.TestAgent testMethod=test_update_device>
    def test_update_device(self):
        """
        Test the methode: update_device of the iota client
        """
    
        device = Device(**self.device)
        device.endpoint = "http://test.com/"
        device.transport = "MQTT"
    
        device.add_attribute(DeviceAttribute(
            name="Att1", object_id="o1", type=DataType.STRUCTUREDVALUE))
>       device.add_attribute(StaticDeviceAttribute(
            name="Stat1", value="test", type=DataType.STRUCTUREDVALUE))
tests/clients/test_ngsi_v2_iota.py:250:
```
The error seems relatively straightforward here: `StaticDeviceAttribute(name="Stat1", value="test", type=DataType.STRUCTUREDVALUE))`. Could you double check again, why this string-like value should have type `STRUCTUREDVALUE`?

tstorek:
@djs0109 I think those are left overs from the initial implementation. Back then `structured_values` were only checked for serialiaziation. Hence, `string`s id not raise errors

examples/ngsi_v2/e11_ngsi_v2_context_specific_models.py
```diff
@@ -0,0 +1,151 @@
...
+    address: AddressAttribute = AddressAttribute()
+
+
+if __name__ == "__main__":
+    # Now we can create the weather station model and export it to a
+    # json-schema file without explicitly defining the entity-type.
+    # Furthermore, we can use the model to create a new weather station entity
+    weather_station = WeatherStation(
+        id="myWeatherStation",
+        type="brick:WeatherStation",
```
djs0109, line 138: I would consider just using "WeatherStation" here, because "brick" is an undefined namespace, and here it is not JSON-LD

examples/ngsi_v2/e11_ngsi_v2_context_specific_models.py
```diff
@@ -0,0 +1,151 @@
...
+
+# Define a context specific model for a weather station
+# First, inherit from the ContextEntity class, where each additional attribute
+# must be of type ContextAttribute
+class WeatherStation(ContextEntity):
+    """
+    A context specific model for a weather station
+    """
+    # add default for type if not explicitly set
+    type: str = FieldInfo.merge_field_infos(
```
djs0109, line 106: What is the difference here to simply use the following code?
```python
    type: Literal["PostalAddress"] = Field(
        # First position is the field info of the parent class
        # set the default value
        default="PostalAddress",
        # overwrite the description
        description="Type of the Address",
        # freeze the field if necessary, e.g. if you want to prevent the user
        # from changing the type after creation.
        frozen=True,
        # for more options see the pydantic documentation
    )
```

filip/models/ngsi_v2/base.py
```diff
@@ -379,17 +420,42 @@ def validate_value_type(cls, value, info: FieldValidationInfo):
...
+            # allows dict and BaseModel as object
+            if type_ == DataType.OBJECT:
+                if isinstance(value, dict):
+                    value = json.dumps(value)
+                    return json.loads(value)
+                elif isinstance(value, BaseModel):
+                    value.model_dump_json()
+                    return value
+                raise TypeError(
+                    f"{type(value)} does not match " f"{DataType.STRUCTUREDVALUE}"
```
djs0109, line 438: DataType.OBJECT ?

djs0109 (changes requested):

djs0109:
@tstorek I made some minor changes in the example, e.g. adding the address into the weather station object. Btw, have you used any formatter? Because I see quit a lot changes on the code format.

tstorek:
> @tstorek I made some minor changes in the example, e.g. adding the address into the weather station object. Btw, have you used any formatter? Because I see quit a lot changes on the code format.

Hi, @djs0109 yes I used black-fromatter. sorry for that. I chekcd on the changes.

examples/ngsi_v2/e11_ngsi_v2_context_specific_models.py
tstorek: That is fine with me

examples/ngsi_v2/e11_ngsi_v2_context_specific_models.py
tstorek: No, it has to be mger_field because otherwise you would overwrite the complete field info from the parent Field, e.g. also patterns etc.

djs0109 (approved):

PR #263, opened 2024-03-26
215 mark quantumleap subscription as deprecated

SystemsPurge:
Marked quantumleap susbscription deprecated as needed. IdPattern was also added as query parameter but was not able to be tested properly.

### DIFF
--- a/tests/clients/test_ngsi_v2_timeseries.py
+++ b/tests/clients/test_ngsi_v2_timeseries.py
@@ -12,7 +12,7 @@ from filip.clients.ngsi_v2 import \
     QuantumLeapClient
 from filip.models.base import FiwareHeader
 from filip.models.ngsi_v2.context import ContextEntity
-from filip.models.ngsi_v2.subscriptions import Message
+from filip.models.ngsi_v2.subscriptions import Message, Subscription
 from filip.utils.cleanup import clean_test, clear_all
 from tests.config import settings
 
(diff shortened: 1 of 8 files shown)


djs0109:
@SystemsPurge, could you add a test function in `tests/clients/test_ngsi_v2_timeseries.py` to validate the implemented `idPattern` functionality?

SystemsPurge:
Test function is implemented and should work properly now. I just wanted to point out a few things. The setup and teardown methods do not work in the test ( at least in test_ngsi_v2_timeseries ) because the servicepath is hard  set on /static, whereas the servicepath in settings ( where it is taken from in the setUp method ) is a different one. No entities are added or removed before and after each test. The test for idPattern is a bit convoluted because in the case where we expect 0 results, an error must be caught, whereas in any other case a value calculated from the response is compared to an expected value. Suggestions on how to write it in a more simplified manner ( if possible ) are definitely welcome.
I've had to go through the quantumleap code to figure out the regex OP used for the SQL queries, and its the '~*' op ( case insensitive POSIX ), same as it stands in ORION documentation, only with an extra ($|.*) tacked on in the end, i guess to indicate an automatic search as a subexpression.

tests/clients/test_ngsi_v2_timeseries.py
```diff
@@ -90,36 +90,27 @@ def test_meta_endpoints(self) -> None:
             self.assertIsNotNone(client.get_version())
             self.assertIsNotNone(client.get_health())
 
-    @clean_test(fiware_service=settings.FIWARE_SERVICE,
```
djs0109: why it should be removed?

djs0109 (changes requested):

tutorials/ngsi_v2/e5_iot_thermal_zone_control/e5_iot_thermal_zone_control_solution.py
djs0109, line 1: Could you revert the changes made for all tutorials? Because it is currently revised in another branch

djs0109 (changes requested):

tests/clients/test_ngsi_v2_timeseries.py
SystemsPurge: If i remember correctly, it is because of the fiware service path. There was a discrepancy between the path in the setup and the path in the settings. I was meaning to put it back.

djs0109 (approved):

PR #264, opened 2024-03-26
262 support entity creation with keyvalues

### LINKED ISSUE #262: Support entity creation with keyvalues
NGSI-V2 supports the creation of data models with `keyValues`, which is not implemented in FiLiP. This feature, however, can benefit the users to reuse use case specific data models, that for example is constructed with `Pydantic`

Implementation of  `keyValues` support for `post_entity()` function. Check this functionality for the function `update()`.

djs0109:
closes #262

### DIFF
--- a/tests/clients/test_ngsi_v2_cb.py
+++ b/tests/clients/test_ngsi_v2_cb.py
@@ -757,9 +757,6 @@ class TestContextBroker(unittest.TestCase):
         entity1_key_value_dict = entity1_key_value.model_dump()
         entity1_key_value_dict["temperature"] = 40
         self.client.update_entity_key_values(entity=entity1_key_value_dict)
-        self.client.get_entity(
-            entity_id=entity1.id,
-            response_format=AttrsFormat.KEY_VALUES).model_dump()
         self.assertEqual(entity1_key_value_dict,
                          self.client.get_entity(
                              entity_id=entity1.id,
--- a/tests/clients/test_ngsi_v2_cb.py
+++ b/tests/clients/test_ngsi_v2_cb.py
@@ -697,6 +697,40 @@ class TestContextBroker(unittest.TestCase):
     @clean_test(fiware_service=settings.FIWARE_SERVICE,
                 fiware_servicepath=settings.FIWARE_SERVICEPATH,
                 cb_url=settings.CB_URL)
+    def test_override_entity_keyvalues(self):
+        entity1 = self.entity.model_copy(deep=True)
+        # initial entity
+        self.client.post_entity(entity1)
+
+        # entity with key value
+        entity1_key_value = self.client.get_entity(
+            entity_id=entity1.id,
+            response_format=AttrsFormat.KEY_VALUES)
+
+        # override entity with ContextEntityKeyValues
+        entity1_key_value.temperature = 30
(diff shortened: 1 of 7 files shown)


djs0109:
@tstorek can you have a look at this? https://github.com/RWTH-EBC/FiLiP/blob/262-Support-entity-creation-with-keyvalues/examples/ngsi_v2/e12_ngsi_v2_use_case_models.py 

Here are some backgrounds of this implementation: In an ongoing research, I have faced the trade-off between the usability of the Pydantic models and the FIWARE- or Filip-compatibility. Therefore, I want to have a way to keep the Pydantic model simple, reusable, while FIWARE- or Filip-compatible.

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -721,7 +756,7 @@ def test_update_entity_keyvalues(self):
         # update entity with dictionary
         entity1_key_value_dict = entity1_key_value.model_dump()
         entity1_key_value_dict["temperature"] = 40
-        self.client.update_entity_key_value(entity=entity1_key_value_dict)
+        self.client.update_entity_key_values(entity=entity1_key_value_dict)
         self.client.get_entity(
             entity_id=entity1.id,
             response_format=AttrsFormat.KEY_VALUES).model_dump()
```
RCX112, line 762: @djs0109 I think this call is redundant

tests/clients/test_ngsi_v2_cb.py
```diff
@@ -961,6 +996,22 @@ def test_batch_operations(self):
...
                                               response_format='keyValues')))
+            # update with keyValues
+            entities_keyvalues = [ContextEntityKeyValues(id=str(i),
+                                                         type=f'filip:object:TypeC',
+                                                         attr1="text attribute",
+                                                         attr2=1
+                                                         ) for i in range(0, 1000)]
+            client.update(entities=entities_keyvalues,
+                          update_format="keyValues",
+                          action_type=ActionType.APPEND)
```
RCX112, line 1007: @djs0109 Is this intended by FIWARE or why can you create multiple entities with the same ID?

djs0109: I think the IDs are different, from 0 to 1000. Or does it show something else in your test?

djs0109:
@RCX112 Thank you for the review, the code has been revised accordingly

tests/clients/test_ngsi_v2_cb.py
RCX112: @djs0109 You can only see that if you look at each specific type separately in the GET request for the OCB endpoint, for example:

`GET http://localhost:1026/v2/entities?limit=1000&type=filip:object:TypeA`

and 

`GET http://localhost:1026/v2/entities?limit=1000&type=filip:object:TypeC`

gives me entities with different types but with the same ID

RCX112: You can also use GET http://localhost:1026/v2/entities?limit=1000&id=0

djs0109: @RCX112 ok, I understand what you mean now. Yes, the IDs of different types are the same. It is fine here, but still thanks for pointing it out

PR #267, opened 2024-04-08
265 order in cleanup function

### LINKED ISSUE #265: order in cleanup function
The order in which the cleanup functions are called is important (afaik, only if there is an iot device holding a command).

In the tutorials, the order of the functions called is correct. 
Yet, we just discovered this when calling the functions in a different order, e.g. clear the cb and then the iota, the iota throws an 400 error that the connection to the cb failed.

Maybe we should add this to the docs somewhere?

SystemsPurge:
Closes #265. Test involves clear methods that are subsequently used by the teardown method, so i added a try , except at the teardown but maybe someone should have a better look into that.

### DIFF
--- a/tests/clients/test_ngsi_v2_iota.py
+++ b/tests/clients/test_ngsi_v2_iota.py
@@ -7,7 +7,7 @@ import logging
 import requests
 
 from uuid import uuid4
-
+from filip.utils.cleanup import clear_iot_agent, clear_context_broker
 from filip.models.base import FiwareHeader, DataType
 from filip.clients.ngsi_v2 import \
     ContextBrokerClient, \
@@ -248,9 +248,9 @@ class TestAgent(unittest.TestCase):
         device.add_attribute(DeviceAttribute(
             name="Att1", object_id="o1", type=DataType.STRUCTUREDVALUE))
         device.add_attribute(StaticDeviceAttribute(
-            name="Stat1", value="test", type=DataType.TEXT))
+            name="Stat1", value="test", type=DataType.STRUCTUREDVALUE))
         device.add_attribute(StaticDeviceAttribute(
-            name="Stat2", value="test", type=DataType.TEXT))
+            name="Stat2", value="test", type=DataType.STRUCTUREDVALUE))
         device.add_command(DeviceCommand(name="Com1"))
 
         # use update_device to post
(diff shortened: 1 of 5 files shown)


tests/clients/test_ngsi_v2_iota.py
```diff
@@ -430,11 +434,30 @@ def test_update_service_group(self):
...
+        while having a device with a command
+
+        Returns:
+            None
+        """
+        device = Device(**self.device)
+        device.add_command(DeviceCommand(name="dummy_cmd"))
+        self.client.post_device(device=device)
+        clear_context_broker(settings.CB_URL,self.fiware_header)
+        self.assertRaises(requests.HTTPError,clear_iot_agent,
```
djs0109, line 449: @SystemsPurge  we prefer to write `assertRaises` this way:
```python
with self.assertRaises(requests.HTTPError):
       clear_iot_agent(settings.IOTA_URL, self.fiware_header)
```
I have commited

tests/clients/test_ngsi_v2_iota.py
```diff
@@ -430,11 +434,30 @@ def test_update_service_group(self):
...
         """
         self.client.close()
-        clear_all(fiware_header=self.fiware_header,
-                  cb_url=settings.CB_URL,
-                  iota_url=settings.IOTA_JSON_URL)
+        try:
+            clear_all(fiware_header=self.fiware_header,
+                    cb_url=settings.CB_URL,
+                    iota_url=settings.IOTA_JSON_URL)
+        except requests.HTTPError:
```
djs0109, line 462: Why must this `except` be added? I think in `clear_all` the oder of the clear functions is correct. And here by capturing all HTTPError, it will show "Already cleared" even if the server is not reachable. Could you refine the check for exception? Otherwise, I would suggest to roll it back

djs0109 (changes requested):

tests/clients/test_ngsi_v2_iota.py
djs0109: @SystemsPurge OK, I understand now. Thanks for explaining. I actually found out that if the registration is deleted, the device can never be deleted via the API ... To make it stable, I would suggest that we can add a flag in clear_context_broker to control whether to clear registrations. By default, we can set it to false, i.e. not to clear regestrations. What do you think?

SystemsPurge: @djs0109 That's convenient, should solve the order issue completely since apparently it only fails to delete a device ( after deleting registration ) if the device has a command.

djs0109: Yes, sounds great :)

djs0109 (approved):

PR #278, opened 2024-04-30
237 update and refactor examples

### LINKED ISSUE #237: Update and refactor examples
Most of the [examples](https://github.com/RWTH-EBC/FiLiP/tree/master/examples) haven't been updated since 2022, so now there's some deprecated code.

Rerun the examples and update any code as necessary. 

**Additional context**
Along the way, typos may be fixed, and examples may be clarified. 

- [x] basic examples
    - [x] refactor
    - [x] fix pydantic bug
- [ ] ngsi_v2
    - [ ] refactor
    - [ ] fix simple_ql bug

**Nice-to-haves**
- [ ] Use colored logs
- [ ] Add descriptions to the log messages as well (in addition to those in the comments)

WaltherTrgovac:
Closes #237.
## Current state of `examples\ngsiv2` that have been worked on:

- Changes in every example:
  - Logger does not show the name of the module and class, but only class
  - Logger does not show milliseconds anymore and I changed the order of day, month, year. Not sure what is the best practise, this can be easily adjusted.
  - Adjusted every example to Python 3.9
 
Before:
Before

After:
after2

Examples:
- `e01_ngsi_v2_context_basics.py`: working ✅ _(last checked: 16.05.2024)_
  - Changes: fixed logger output, entity ids changed to the standard ones
- `e02_ngsi_v2_context_relationships.py`: working ✅ _(last checked: 24.04.2024)_
- `e03_ngsi_v2_context_subscriptions_http.py`: working ✅ _(last checked: 21.05.2024)_
  - Changes: reformatted subscription comment
- `e04_ngsi_v2_context_subscriptions_mqtt.py`: working ✅ _(last checked: 21.05.2024)_
  - Changes: fixed deprecated methods, reformatted mqtt notifications, removed unused imports
- `e05_ngsi_v2_context_registrations.py`: not completely working ⚠️  _(last checked: 29.05.2024)_
  - Changes: fixed deprecated methods, PEP-8 errors, and reformatted comments.
  - Notes: It seems as if the context provider weather station is not providing any data to the building that is supposed to be provided from the weather station when we query the building at the end of the example
- `e06_ngsi_v2_autogenerate_context_data_models.py`: not working ❌  _(last checked: 30.04.2024)_
  - Notes: A very short example, but half of the code is deprecated, maybe it should be completely remade?
- `e07_ngsi_v2_iota_basics.py`: not completely working ⚠️  _(last checked: 29.05.2024)_
  - Changes: fixed deprecated methods, typos, and pep-8 errors. Changed entity ids to the standard ones.
  - Notes: The code gets executed, but with one small error from logger: `2024-04-30 17:03:55,412 filip.clients.ngsi_v2.cb.ContextBrokerClient ERROR: Could not delete attribute 'on' of entity'sensor2'  
 Reason: {"error":"NotFound","description":"The requested entity has not been found. Check type and id"}`
- `e08_ngsi_v2_iota_paho_mqtt.py`: working ✅  _(last checked: 11.06.2024)_
  - Changes: fixed deprecated methods, pep 8 errors, typos. Also fixed the error with random attribute values being posted differently in comparison to the logger output. Fixed QoS message from mqtt client.
- `e09_ngsi_v2_iota_filip_mqtt.py`: working: ✅   _(last checked: 12.06.2024)_
  - Changes: fixed deprecated methods, adjusted mqtt callback function logger statements, adjusted temperature value to be in 'normal' range.
  - Note 📓 : This example is showing more just the capabilities of the library, but not on the actual use case. Some methods are just listed in the example, so it is a bit messy and not completely clear what the example wants to show. This example can be maybe structured better?
- `e10_ngsi_vs_quantumleap.py`: working ✅ _(last checked: 07.05.2024)_
  - Changes: fixed PEP-8 errors, typos. Removed milliseconds from logger for better readability. Changed print statements for better readability. 
  - Notes: `clear_all()` method at the start is causing this message from the logger: `filip.clients.ngsi_v2.quantumleap.QuantumLeapClient ERROR: Could not load entity data`
 `Reason: {
  "description": "No records were found for such query.",
  "error": "Not Found"
}` This does not cause any negative impacts further in the code, but the output is not very clean.
- `e11_ngsi_v2_context_specific_models.py`: working ✅ _(last checked: 08.05.2024)_
  - Changes: None
- `e12_ngsi_v2_use_case_models.py`: working ✅ _(last checked: 18.06.2024)_
  - Changes: clear_context_broker() method was missing parameters, now they have been added.

TODO: The directories of the examples and the numeration of the examples has to be adjusted. For example, we have two examples with the number 11 in the `ngsiv2` directory, and the examples in the `basics` directory also need adjustment. README.md then also needs to be adjusted accordingly.


examples/ngsi_v2/e08_ngsi_v2_iota_paho_mqtt.py
Maghnie, line 1: For some unknown reason, this example doesn't work for me. The temperature and humidity values are “null” after all the for-loop iterations.

examples/ngsi_v2/e09_ngsi_v2_iota_filip_mqtt.py
Maghnie, line 1: This example still produces warnings. Can they be avoided?

examples/ngsi_v2/e10_ngsi_v2_quantumleap.py
Maghnie, line 1: There are many nested quotation marks of the same type here.

Maghnie (changes requested):
Thanks for the extensive initial updates!

examples/ngsi_v2/e01_ngsi_v2_context_basics.py
```diff
@@ -43,8 +43,8 @@
                                     fiware_header=fiware_header)
     # View version
     for key, value in cb_client.get_version().items():
-        logger.info("Context broker version" + value["version"] + " at url " +
-                    cb_client.base_url)
+        logger.info(f"Context broker version: {value["version"]} at url: "
```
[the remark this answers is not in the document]
WaltherTrgovac: Fixed. Python version 3.9 does not allow nested quotation marks of the same type. Now every example can also be executed in Python 3.9.

examples/basics/e01_http_clients.py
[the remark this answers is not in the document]
WaltherTrgovac: I think it does actually read it, because if you try to change something in the env file, you will get errors in the example.

examples/ngsi_v2/e10_ngsi_v2_quantumleap.py
WaltherTrgovac: Fixed.

examples/ngsi_v2/e09_ngsi_v2_iota_filip_mqtt.py
WaltherTrgovac: This is happening due to the nature of the example. We could just comment out the part that is producing the error. This is the part 3.2 in this case, but I think the point of the example is to show the methods and not an actual real case usage, but I am not sure.

djs0109 (changes requested):
Now there is some problem with the reading of `.env.filip` . It seems like the variables will only be correctly loaded when the `.env.filip` is put under the path of the examples, i.e. `examples/ngsi_v2/` or `examples/basics/` @WaltherTrgovac could you have a look on that? If necessary you can modify the `filip/config.py`, where the variables are loaded

Maghnie:
> Now there is some problem with the reading of `.env.filip` . It seems like the variables will only be correctly loaded when the `.env.filip` is put under the path of the examples, i.e. `examples/ngsi_v2/` or `examples/basics/` @WaltherTrgovac could you have a look on that? If necessary you can modify the `filip/config.py`, where the variables are loaded

Wasn't it always like that? I think in the config of filip it says that the env file should be in the working directory, so it's up to the user to take care of that.

WaltherTrgovac:
> Now there is some problem with the reading of `.env.filip` . It seems like the variables will only be correctly loaded when the `.env.filip` is put under the path of the examples, i.e. `examples/ngsi_v2/` or `examples/basics/` @WaltherTrgovac could you have a look on that? If necessary you can modify the `filip/config.py`, where the variables are loaded

If I am understanding this correctly, the method `SettingsConfigDict` in the class `Settings` is taking the path of the `.env.filip` file. It was set so that it is found only when the files to execute are in the same directory as the `.env.filip`. 
I have now adjusted that path so that the file can be located anywhere in the repository. I use the glob python package for that. First I locate the path of the repository with GitPython package and use that as the root directory. Then I look through the whole repository for the `.env.filip` file. The problem is that the `.env.filip` is a hidden file and glob package supports searching for hidden files only from the Python version 3.11... 😞. If somebody knows a solution for the earlier versions of Python that would be great.
And another thing, I have no idea how to test this implementation, I have just tested it manually...

djs0109:
> If somebody knows a solution for the earlier versions of Python that would be great.

Never mind, I have changed it now

Maghnie:
> > If somebody knows a solution for the earlier versions of Python that would be great.
> 
> Never mind, I have changed it now

Then to avoid confusion between multiple env files, we should probably remove the copy under [basics](https://github.com/RWTH-EBC/FiLiP/tree/237-update-and-refactor-examples/examples/basics). Or do you see another use for it?

djs0109:
Nice hint. No we don't need it any longer

Maghnie (approved):

djs0109:
Now only e6 cannot work, because the code regarding data model creation is not yet migrated to pydantic V2

djs0109 (approved):

PR #279, opened 2024-04-30
178 error on duplicate object

### LINKED ISSUE #178: Error on duplicate object_id
If a device as two attrs with the same object_id, fiware would randomly send the value to only one attr.

Warn / throw errors in that case for a device-config.

djs0109:
closes #178

### DIFF
--- a/tests/models/test_ngsi_v2_iot.py
+++ b/tests/models/test_ngsi_v2_iot.py
@@ -214,6 +214,70 @@ class TestContextv2IoTModels(unittest.TestCase):
 
             assert len(w) == 2
 
+    def test_device_attributes(self):
+        """
+        Test the device model regarding the behavior with devices attributes.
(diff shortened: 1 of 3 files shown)


tests/models/test_ngsi_v2_iot.py
```diff
@@ -214,6 +214,142 @@ def test_expression_language(self):
...
+        attr1 = DeviceAttribute(
+            name="temperature",
+            type="Number"
+        )
+        with self.assertRaises(ValueError):
+            device_a.add_attribute(attribute=attr1)
+
+        # fail, because the object_id is duplicated with the name of attr1
+        device_b = initial_device()
+        attr2 = DeviceAttribute(
```
sbanoeon, line 249: The name must not be unique, right? Why is object_id compared with name of attr1?

sbanoeon (approved):

PR #293, opened 2024-06-25
198 support of alterationtypes in subscriptions

### LINKED ISSUE #198: Support of `alterationTypes` in subscriptions
**Descriptions**
`alterationTypes` is added as new feature of `Subscriptions` since ~~Orion 3.8~~ Orion 3.7, see [here](https://github.com/telefonicaid/fiware-orion/blob/3.8.0/doc/manuals/orion-api.md#subscriptionsubjectcondition).

That allows the activation of subscriptions based on the following settings:
- `entityUpdate`: notification is sent whenever a entity covered by the subscription is updated (no matter if the entity actually changed or not)
- `entityChange`: notification is sent whenever a entity covered by the subscription is updated and it actually changes (or if it is not an actual update, but forcedUpdate option is used in the update request)
- `entityCreate`: notification is sent whenever a entity covered by the subscription is created
- `entityDelete`: notification is sent whenever a entity covered by the subscription is deleted

**Use case**
There can be many use cases. One potential use case is, using `entityCreate` to trigger some workflows (e.g. generating Grafana dashboard) every time a new entity is created. 

**Implementation**
The data model of `subscriptions` need to be extended, i.e. `subscriptions.subject.condition.alterationTypes`

djs0109:
closes #198

### DIFF
--- a/filip/models/ngsi_v2/subscriptions.py
+++ b/filip/models/ngsi_v2/subscriptions.py
@@ -274,6 +274,8 @@ class Condition(BaseModel):
     expression matches.
     If neither attrs nor expression are used, a notification is sent whenever
     any of the attributes of the entity changes.
+    alterationTypes: for more information about this field, see
+    https://github.com/telefonicaid/fiware-orion/blob/3.8.0/doc/manuals/orion-api.md#subscriptions-based-in-alteration-type
 
     """
     attrs: Optional[Union[str, List[str]]] = Field(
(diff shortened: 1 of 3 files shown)


filip/models/ngsi_v2/subscriptions.py
```diff
@@ -296,6 +302,20 @@ def check_attrs(cls, v):
...
+    @field_validator('alterationTypes')
+    def check_alteration_types(cls, v):
+        allowed_types = {"entityCreate", "entityDelete", "entityUpdate", "entityChange"}
+
+        if v is None:
+            return None
+        elif isinstance(v, list):
+            for item in v:
+                if item not in allowed_types:
+                    raise ValueError(f'{item} is not a valid alterationType')
```
sbanoeon, line 314: It would be super helpful if this error message also list the allowed alterationTypes.

sbanoeon (approved):

filip/models/ngsi_v2/subscriptions.py
djs0109: Sure. I have implemented it

PR #295, opened 2024-06-27
Adding new attributes to the HttpCustom model 

RCX112:
Extended the pydantic model with the attributes json, ngsi and timeout. Also added some tests for the pydantic model.

### DIFF
--- a/filip/clients/ngsi_v2/cb.py
+++ b/filip/clients/ngsi_v2/cb.py
@@ -1898,7 +1898,7 @@ class ContextBrokerClient(BaseHttpClient):
         except requests.RequestException as err:
             msg = (
                 f"Sending notifcation message failed! \n "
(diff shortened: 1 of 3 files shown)


djs0109:
@RCX112 I think there might be some misunderstanding. The issue #290 should actually add `headers` under `httpCustom`. The related docs are [here](https://fiware-orion.readthedocs.io/en/3.8.0/orion-api.html#custom-notifications)

djs0109:
OK, I get it now. The `headers` already exists

djs0109:
@RCX112 I will then merge #290  to #288 and then close the #290 . What do you think?

RCX112:
@djs0109 Sorry for the misunderstanding, we discussed this problem also in yesterday's meeting. I think merging both branches is a good idea

djs0109:
Close, because it is merged to #288

PR #296, opened 2024-07-01
288 filip support right now only paylaod in mqtt custom / http custom but orion already supports json and ngsi as further options

SystemsPurge:
The following warning pops up "UserWarning: Field name "json" shadows an attribute in parent "Mqtt"" ( due to the json fields in MqttCustom and HttpCustom ). Since the Http/Mqtt models do not have an explicit json field, i take it this is coming from some field/property in the base model.
Tried to circumvent the warning by the use of aliases which obviously did not work.

Not sure what model to specify for the json payload type, i just put a Dict[str, any] for now ( despite number/string being valid json , the orion manual specifies that it needs to be a JSON object and not an object of primitive type )

Any input on these two points is welcome.

### DIFF
--- a/tests/models/test_ngsi_v2_subscriptions.py
+++ b/tests/models/test_ngsi_v2_subscriptions.py
@@ -145,7 +145,8 @@ class TestSubscriptions(unittest.TestCase):
         Returns:
             None
         """
-        sub = Subscription.model_validate(self.sub_dict)
+        tmp_dict=self.sub_dict.copy()
+        sub = Subscription.model_validate(tmp_dict)
         fiware_header = FiwareHeader(service=settings.FIWARE_SERVICE,
                                      service_path=settings.FIWARE_SERVICEPATH)
         with ContextBrokerClient(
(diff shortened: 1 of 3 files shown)


tests/models/test_ngsi_v2_subscriptions.py
djs0109, line 1: Could you please also add tests that send at least 3 subscriptions (payload, json, ngsi) to Orion endpoints and then validate them by querying the created subcriptions from Orion? The idea is to make sure the created data models are also valid for Orion

djs0109 (changes requested):

djs0109:
closes #288

djs0109:
I think it would be great if we can test the functionality of these fields too. Could you please implement a test that uses `json`, `ngsi`, `payload` for `mqttCustom`? An example can be found here https://github.com/RWTH-EBC/FiLiP/blob/7919664cad1d7ff3800724f819a7c9f43bd28a80/tests/clients/test_ngsi_v2_cb.py#L832

RCX112:
> The following warning pops up "UserWarning: Field name "json" shadows an attribute in parent "Mqtt"" ( due to the json fields in MqttCustom and HttpCustom ). Since the Http/Mqtt models do not have an explicit json field, i take it this is coming from some field/property in the base model. Tried to circumvent the warning by the use of aliases which obviously did not work.
> 
> Not sure what model to specify for the json payload type, i just put a Dict[str, any] for now ( despite number/string being valid json , the orion manual specifies that it needs to be a JSON object and not an object of primitive type )
> 
> Any input on these two points is welcome.

@SystemsPurge When I implemented the HttpCustom model I also had the same warnings. I simply supressed them with the warnings library, because the json() method of pydantic is already outdated anyway:

```
import warnings

# The pydantic models still have a .json() function, but this method is deprecated.
warnings.filterwarnings("ignore", category=UserWarning,
                        message='Field name "json" shadows an attribute in parent "Http"')
warnings.filterwarnings("ignore", category=UserWarning,
                        message='Field name "json" shadows an attribute in parent "Mqtt"')
```

RCX112:
@djs0109 Which version of Orion do we currently use for the tests, because the custom notifications don't work for version 3.7 for example?

djs0109:
Hi @RCX112 , it should be >= 3.8

djs0109 (approved):
@SystemsPurge @RCX112 LGTM, thank you!

PR #320, opened 2024-09-03
301 ngsi ld validate nested properties dynamically

Maghnie:
Properties are now checked dynamically/recursively. 

But list properties appear to cause errors:

```
======================================================================
ERROR: test_validate_subproperties_list (tests.models.test_ngsi_ld_context.TestLDContextModels)
Test the validation of multi-level properties in entities
----------------------------------------------------------------------
Traceback (most recent call last):
  File "d:\Repos\04-OpenSource\FiLiP\tests\models\test_ngsi_ld_context.py", line 277, in test_validate_subproperties_list
    entity4 = ContextLDEntity(**self.entity_list_property)
  File "d:\Repos\04-OpenSource\FiLiP\filip\models\ngsi_ld\context.py", line 503, in __init__
    data.update(self._validate_attributes(data))
  File "d:\Repos\04-OpenSource\FiLiP\filip\models\ngsi_ld\context.py", line 523, in _validate_attributes
    attrs[key] = ContextProperty.model_validate(attr)
  File "C:\Users\mmg\AppData\Local\anaconda3\envs\filip\lib\site-packages\pydantic\main.py", line 503, in model_validate
    return cls.__pydantic_validator__.validate_python(
pydantic_core._pydantic_core.ValidationError: 1 validation error for ContextProperty
  Input should be a valid dictionary or instance of ContextProperty [type=model_type, input_value=[{'type': 'Property', 'va...erty:gpsBxyz123-speed'}], input_type=list]
    For further information visit https://errors.pydantic.dev/2.5/v/model_type
```


djs0109:
As discussed yesterday, we would like to first not support the list property, coming from different sources

Maghnie:
The recursive validation needed a fix, which is done now.

I added two new unit tests for testing the validation of subproperties. One test uses an entity dict with valid subproperties and another uses an invalid dict:
image

Both validations pass successfully. However, for the invalid dict, the wrong property declarations are automatically "fixed" by filip:

image

The question here is, whether this behavior is suitable, or is it better to simply reject the entity creation and request users to fix the property declarations themselves. If we're going with option 1, then this issue can be resolved.

djs0109:
@Maghnie looks good! Regarding the preferred behavior you mention, the option 2 makes more sense to me. Otherwise the users may keep the error in their script of data.

djs0109:
@Maghnie Could you look into this PR again? I have merged a branch and changed quit a lot. If that looks fine to you, I will merge and close this PR 🙂

Maghnie:
@djs0109 Thanks for taking care of the issue during my vacation :) 

The changes look good to me! especially the part with not having to check for geoproperty first :D

djs0109 (approved):

PR #329, opened 2024-09-24
325 filip mqtt client still require object

### LINKED ISSUE #325: Filip MQTT Client still require object_id
**Describe the bug**
The `object_id `is used in the iotagent_mqtt_client. However, it is not a required variable for device attribute. The iotgent_mqtt_client needs to be adapted

SystemsPurge:
Closes #325 . Test in test_mqtt_client is practically standalone and should not interfere with the other tests

### DIFF
--- a/filip/clients/mqtt/client.py
+++ b/filip/clients/mqtt/client.py
@@ -751,13 +751,22 @@ class IoTAMQTTClient(mqtt.Client):
                 msg_payload = payload.copy()
                 for key in payload.keys():
                     for attr in device.attributes:
-                        if key in attr.object_id or key == 'timeInstant':
+                        key_constraint = key == "timeInstant"
+                        def elif_action(msg): None
+
+                        if attr.object_id is not None:
+                            key_constraint = key_constraint or (key in attr.object_id)
+                            def elif_action(msg): msg[attr.object_id] = msg.pop(key)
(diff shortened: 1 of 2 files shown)


djs0109:
@SystemsPurge do you think it make sense to use `msg_payload` instead of `payload` in line781?
https://github.com/RWTH-EBC/FiLiP/blob/f99807e6b18d9aff6ee0d5f20aa23621228097bd/filip/clients/mqtt/client.py#L779C1-L782C35
I thought they should be the same because of the shallow copy. But it is not

SystemsPurge:
> @SystemsPurge do you think it make sense to use `msg_payload` instead of `payload` in line781? https://github.com/RWTH-EBC/FiLiP/blob/f99807e6b18d9aff6ee0d5f20aa23621228097bd/filip/clients/mqtt/client.py#L779C1-L782C35

To be perfectly honnest, i don't know 100% what the intended use is under the hood, however, it is clear that , at least in the first case in the publish function (attr_name None and command None), the payload is being copied to be manipulated ( e.g popping the attr name key and replacing it with the object_id key ), but then it is not passed anywhere, so it would make sens to pass it there instead of payload. 
TL;DR : Given how the code is written, yes it makes sens to pass message_payload instead of payload

PR #335, opened 2024-10-22
305 detailed exception handling inside different filip methods

RCX112:


RCX112:
@mwr-ebc Some tests are missing because there are a lot of different requests inside the clients and I didn't have the time to add all of them. Can you maybe help me with that?

mwr-ebc (approved):

PR #354, opened 2024-11-19
chore: include pre-commit configuration

sbanoeon:


djs0109:
I saw that pre-commit is creating dependency problem with py38. Maybe we can include it as an optional dependency under `extras_require`

djs0109:
It seems like some of the pre-configured timeseries data is missing. I will have a check

djs0109:
PR move into https://github.com/RWTH-EBC/FiLiP/pull/364

PR #388, opened 2025-02-08
315 unittest via GitHub actions

Bhavdeep13:
Workflow CI for FiLiP
- includes workflow with secret variables for env file
- workflow.md file for documentation


djs0109:
- [ ] @djs0109 check the `checkout code` part in the gitub workflow, whether it is working fine

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,99 @@
...
+          cd filip_unittest/docker
+          docker compose up -d
+
+      # Step 7: Wait for FIWARE Services
+      - name: Wait for FIWARE Services
+        run: |
+          for i in {1..30}; do
+            curl -s http://localhost:1026/version && \
+            curl -s http://localhost:4041/iot/about && \
+            curl -s http://localhost:8668/version && break || sleep 5
```
sbanoeon, line 74: check endpoint of Orion LD?

sbanoeon (changes requested):

.gitlab-ci.yml
```diff
@@ -7,7 +7,7 @@ stages:
     - deploy
 
 variables:
-    EXCLUDE_PYTHON: 37
+    EXCLUDE_PYTHON: 37, 38, 39, 310, 311
```
sbanoeon, line 10: Where is this variable used?

.github/workflows/unittest.yml
djs0109: Good point, it should be fixed now

.gitlab-ci.yml
djs0109: In this case, the unittest in Gitlab will only test for 312 and coverage, so that the pipline will run faster

sbanoeon (approved):

sbanoeon:
Could you also please edit Changelog file.

PR #390, opened 2025-02-12
chore: relax pydantic requirements

### LINKED ISSUE #387: Update dependencies
Hi,

it would be great if you could relax the requirements for some packages, e.g. Pydantic (now 2.7.0) which still has some annoying bugs in the alias system. https://github.com/pydantic/pydantic/pull/9561 
Things move fast. It would be great if the dependencies would be checked on regular basis.
That would favor better debug capabilities through updating packages.

Thanks in advance

djs0109:
closes #387


djs0109:
There is an update of the `create_model` since pydantic 2.8, which break the fucntion `create_context_entity_model` in filip. We need to make sure after the dependency relax, our codes work for both pydantic <2.8 and ==2.8

tstorek:
@djs0109 Could you please elaborate on the error. I am pretty packed right now, but maybe we can check next week.
From a quick glance the type-hint should not impact the model creation. Most likely it is only a small thing. :)

djs0109:
In ``pydantic 2.8`` They changed 
```python
def create_model(  # noqa: C901
    __model_name: str,
    *,
...
```
to 
```python
def create_model(  # noqa: C901
    model_name: str,
    *,
...
```
So I have to change to use a positional argument to pass the `model_name`. It is not the most elegant solution, but it should work

PR #393, opened 2025-02-17
chore: include context conversion methods and test

sbanoeon:


tests/models/test_ngsi_v2_context.py
```diff
@@ -623,6 +623,12 @@ def test_get_attributes(self):
         self.assertNotEqual(entity.get_attributes(strict_data_type=True), attributes)
         self.assertNotEqual(entity.get_attributes(), attributes)
 
+    def test_context_conversion(self):
+        entity_normalized = ContextEntity(**self.entity_data)
+        entity_key_values = ContextEntityKeyValues(**self.entity_data)
```
djs0109: I think this way the created `entity_key_values` is not really in keyvalue format.
Expected `entity_key_values.model_dump_json()`
```json
{"id": "MyId", "type": "MyType", "temperature": 20, "relation": "OtherEntity"}
```
Right now
```json
{"id": "MyId", "type": "MyType", "temperature": {"value": 20, "type": "Number"}, "relation": {"value": "OtherEntity", "type": "Relationship"}}
```

djs0109: I made adjustments accordingly. However, the test is failing, so the conversion functionality needs to be checked again.

djs0109 (changes requested):

tests/models/test_ngsi_v2_context.py
sbanoeon: Hmm, now I am thinking, shouldn't the constructor of `ContextEntityKeyValue` class take care of the right format?

djs0109:
The failing tests are related to https://github.com/RWTH-EBC/FiLiP/pull/407. After pulling it should work inmediately

djs0109 (approved):

PR #394, opened 2025-02-18
384 ngsi ld support for entity creation and update with keyvalues format

SystemsPurge:
There might still be some redundancies

### DIFF
--- a/filip/clients/ngsi_ld/cb.py
+++ b/filip/clients/ngsi_ld/cb.py
@@ -196,7 +196,10 @@ class ContextBrokerLDClient(BaseHttpClient):
             raise
 
     def post_entity(
-        self, entity: ContextLDEntity, append: bool = False, update: bool = False
+        self, 
+        entity: Union[ContextLDEntity,ContextLDEntityKeyValues], 
+        append: bool = False, 
+        update: bool = False
     ):
         """
         Function registers an Object with the NGSI-LD Context Broker,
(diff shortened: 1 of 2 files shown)


tests/clients/test_ngsi_ld_cb_kv.py
```diff
@@ -0,0 +1,998 @@
...
+        Returns:
+            None
+        """
+        self.resources = {
+            "entities_url": "/ngsi-ld/v1/entities",
+            "types_url": "/ngsi-ld/v1/types",
+        }
+        self.attr = {"testtemperature": {"type": "Property", "value": 20.0}}
+        self.entity = ContextLDEntityKeyValues(
+            id="urn:ngsi-ld:my:id4", type="MyType", **self.attr
```
djs0109, line 51: The entity created in this way is not really in keyvalues format. The attributes should have format like:
```json
{"testtemperature": 20.0}
```

filip/clients/ngsi_ld/cb.py
```diff
@@ -524,7 +527,7 @@ def update_entity_attribute(
             raise
 
     def append_entity_attributes(
```
djs0109, line 529: I am not sure whether keyvalues format is supported in this endpoint `.../entities/{entity.id}/attrs`

djs0109: I will check it out

tests/clients/test_ngsi_ld_cb_kv.py
SystemsPurge: i will change everything accordingly and test again @djs0109

filip/clients/ngsi_ld/cb.py
SystemsPurge: @djs0109 Patching attributes certainly requires a pydantic basemodel inheriting class because the method attemps to model_dump() the attributes to patch... Passing a dictionary causes it to throw

SystemsPurge: @djs0109 i removed the patch attrs test for now for that reason

djs0109 (approved):

PR #407, opened 2025-03-10
chore: add time sleep to increase stability

djs0109:


mwr-ebc (approved):
For LD this change is working fine.

PR #417, opened 2025-04-15
remove irrelevant params for get_device_list

### LINKED ISSUE #323: IoTaC.get_device_list add unrelevant data to request parameters
**Describe the bug**
With the following code, unrelevant data will as parameter with the request. And the actual relevant `limit` and `offset` are not being sent. 
```python
        params = {key: value for key, value in locals().items() if value is not
                  None}
```

**To Reproduce**
Add a breakpoint here https://github.com/RWTH-EBC/FiLiP/blob/70006d6c327e574c5c83aa7d9a7892a0e30764f5/filip/clients/ngsi_v2/iota.py#L358
and run `test_ngsi_v2_iota.py` in debug mode and check the `params`

djs0109:
closes #323

### DIFF
--- a/filip/clients/ngsi_v2/iota.py
+++ b/filip/clients/ngsi_v2/iota.py
@@ -380,13 +380,21 @@ class IoTAClient(BaseHttpClient):
         Returns:
             List of matching devices
         """
+        params = {}
         if limit:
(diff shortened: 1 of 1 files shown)


mwr-ebc (approved):
everythings looks fine.

PR #432, opened 2025-06-03
NGSI LD examples

SystemsPurge:
I have unfortunately wasted a lot of time trying to figure out how to circumvent the custom model property type issues (e06 use case models), and have only annotated the examples i wrote from scratch. I've also felt that the JSON-LD context example is a bit on the light side so i added an example about multi-tenancy.
You will notice that there are some source code files that have been modified, this is to make the quantumleap example work with ld, as a model for a manual QL notification does not exist in ld, so i added it.


examples/ngsi_ld/e09_ngsi_ld_sysattrs.py
```diff
@@ -0,0 +1,100 @@
+"""
+Usually the live data is stored in Context Broker and the historical data in the time
```
djs0109, line 2: The description is not relevant for this example

PR #513, opened 2026-08-03
fix: datatime validation in metadata

### LINKED ISSUE #509: Inconsisten Serialization behaviour
**Describe the bug**
It is not possible to create `Metadata` of a `ContextAttribute` (NGSIv2) of type `datetime.datetime`, even if the default pydantic seriliazer would support it

**To Reproduce**
A reduced implementation of metadata looks as follow
```python
class Metadata(BaseModel):
    value: Any
    type: Optional[Union[DataType, str]] = Field(default=None)

    @field_validator("value")
    def validate_value(cls, value, info: ValidationInfo):
       assert json.dumps(value), "metadata not **serializable**
```
If I try to create an according `Metadata` instance with
```
data = {
    "type": "DateTime",
    "value": datetime.datetime.now(tz=datetime.UTC)
}
Metadata.model_validate(data)
```
i get `TypeError: Object of type datetime is not JSON serializable` even if in theory pydantic could serialize the value.

The problem stems from this validator
https://github.com/RWTH-EBC/FiLiP/blob/27c7d3162829548a4c2f0cf03092c46f6c614f28/filip/models/ngsi_v2/base.py#L229-L235

Here the serializability is checked using the builtin json library and an assertion. In addition to the differing functionality of `json` and `pydantic` in regards to serialization, the assertion would be skipped either way if I run python using the `-O` parameter.

I would propose switchting the validator of `Metadata` so it looks something like
```python
class Metadata(BaseModel):
    value: Any
    type: Optional[Union[DataType, str]] = Field(default=None)

    @model_validator(mode="after")
    def validate_serializability(self, info: ValidationInfo):
        self.model_dump(mode="json")
```

**Environment (please complete the following information):**
 - OS: Windows 10 (problem persists in WSL, Ubuntu-24.04)
 - Filip: v0.7.5
 - Pydantic: 2.8.3
 - Python: 3.12 (but persists through 3.11+)

**Additional context**
The whole code I used for recreating the behaviour
```python
import datetime
import json
import logging
from typing import Optional, Union, Any

from filip.models.base import DataType
from pydantic import BaseModel, Field, model_validator, field_validator
from pydantic_core.core_schema import ValidationInfo

logging.basicConfig(level="INFO")
_logger = logging.getLogger("Main")


class Metadata(BaseModel):
    value: Any
    type: Optional[Union[DataType, str]] = Field(default=None)

    # @field_validator("value")
    # def validate_value(cls, value, info: ValidationInfo):
    #     assert json.dumps(value), "metadata not serializable"

    @model_validator(mode="after")
    def validate_serializability(self, info: ValidationInfo):
        self.model_dump(mode="json")


if __name__ == "__main__":
    data = {
        "type": "DateTime",
        "value": datetime.datetime.now(tz=datetime.UTC)
    }
    Metadata.model_validate(data)
```
Just switch between the two version of validator to reproduce the bahaviour

djs0109:
closes #509

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 - update: deprecate support for python 3.9 ([#504](https://github.com/RWTH-EBC/FiLiP/pull/504))
 - update: from ``requests~=2.32.0`` to ``requests~=2.33.0`` in dependency ([#504](https://github.com/RWTH-EBC/FiLiP/pull/504))
 - fix: exception handling in checking broker version ([#503](https://github.com/RWTH-EBC/FiLiP/pull/503))
+- fix: datatime validation in metadata ([#513](https://github.com/RWTH-EBC/FiLiP/pull/513))
+
 
 ### v0.7.5
 - fix: allow dynamic header composition in v2 http client ([#493](https://github.com/RWTH-EBC/FiLiP/pull/493))
--- a/filip/models/ngsi_v2/base.py
+++ b/filip/models/ngsi_v2/base.py
@@ -28,6 +28,7 @@ from pydantic import (
 )
 
 from typing import Union, Optional, Pattern, List, Dict, Any
+from typing_extensions import Self
 
 from filip.models.base import DataType
 from filip.models.ngsi_v2.units import validate_unit_data, Unit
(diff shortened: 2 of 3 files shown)


sjohnen (approved):
lgtm

The failing tests for Python 3.10 could be fixed by creating the metadata timestamp from `datetime.fromisoformat('2011-11-04 00:05:23.283+00:00')` instead of `datetime.datetime.now(tz=datetime.UTC)`, since `datetime.UTC` was introduced in Python 3.11

